PR #2, opened 2025-12-04
Bash Pocket

hakangoksu:
I made a plugin named bash pocket. 

This is a plugin to have multiple custom bash scripts to execute/edit easily on using the main bar.

Purpose: You are using spesific couple of commands and instead of going different terminals or copying pasting same stuff again and again, you can use this plugin.

Example: You are developing something in a folder. You want to execute it with certain commands or you want to copy them with bash command to a live server/location. You can save some commands in bash pockets and you can click on the pocket to execute it easily.

This basically has limitless usage since it is direct support for bash scripts on the bar.

I tested and seems to be working well on the latest shell.


bash-pocket/BarWidget.qml
```diff
@@ -0,0 +1,282 @@
...
+                refreshPockets();
+                editPocket(newIndex);
+            }
+        }
+    }
+
+    function editPocket(index) {
+        var path = root.pluginDir + "/" + index + ".sh";
+        
+        // If $EDITOR is set, use it; otherwise use kitty -e nano
```
Ly-sec: I don't think you should default to kitty, there is no need to define any terminal. You can always either use Process with sh or something else.

bash-pocket/BarWidget.qml
```diff
@@ -0,0 +1,282 @@
...
+            });
+            model.push({
+                "label": "Delete " + index,
+                "icon": "trash",
+                "action": "delete_" + index
+            });
+            // Add separator if not last
+            if (i < root.pockets.length - 1) {
+                 // NContextMenu doesn't support separators natively in the model structure used here easily without custom delegate logic or model types, 
+                 // but we can just list them sequentially.
```
Ly-sec: Clean up AI comments :p

bash-pocket/BarWidget.qml
```diff
@@ -0,0 +1,282 @@
...
+            border.width: Style.borderS
+            radius: Style.radiusM
+            
+            ColumnLayout {
+                id: contentLayout
+                width: parent.width
+                anchors.centerIn: parent
+                spacing: Style.marginS
+                
+                // Header / Title (Optional, but good for spacing)
```
Ly-sec: Same here

hakangoksu:
First of all thank you very much for the interest! I'm very happy you find this interesting enough.

I will delete AI comments I'm trying this Antigravity super usefull automatically understands what you want to do even manually coding, suggestions are amazing (without the agent even)

About the fallback, if I don't do a fallback (eg. kitty nano) it doesn't seem to open anything for me.

Any suggestions for fallback? I couldn't come up with a better idea for this.

ItsLemmy:
also missing a preview.png

hakangoksu:
> also missing a preview.png

Added.

For editor fallback, I'm open for ideas.

PR #279, opened 2026-02-19
feat: initial version of projection settings

Sparkenstein:
Adds a simple plugin for multi monitor setup that you can just click to switch the profile. 

image

Needs `kanshi`.

PS: I don't know any qml, this is the best I could manage, let me know any improvements

### DIFF
--- a/projection-settings/BarWidget.qml
+++ b/projection-settings/BarWidget.qml
@@ -1,0 +1,44 @@
+import QtQuick
+import QtQuick.Layouts
+import Quickshell
+import qs.Commons
+import qs.Widgets
+
+Rectangle {
+    id: root
+    
+    property var pluginApi: null
+    property ShellScreen screen
+    property string widgetId: ""
+    property string section: ""
+
+    implicitWidth: Style.barHeight
+    implicitHeight: Style.barHeight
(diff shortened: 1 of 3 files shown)


projection-settings/BarWidget.qml
```diff
@@ -0,0 +1,44 @@
...
+
+Rectangle {
+    id: root
+    
+    property var pluginApi: null
+    property ShellScreen screen
+    property string widgetId: ""
+    property string section: ""
+
+    implicitWidth: Style.barHeight
```
ItsLemmy: This is outdated, please use the multi-monitor ways, check "hello-world" plugin

projection-settings/BarWidget.qml
```diff
@@ -0,0 +1,44 @@
+import QtQuick
+import QtQuick.Layouts
+import Quickshell
+import qs.Commons
+import qs.Widgets
+
+Rectangle {
```
ItsLemmy: I believe overall NIconButton would be better here, than a custom implementation via Rectangle.
This will look better and will support propert capsule color, outline, etc...

ItsLemmy:
missing the readme and the preview.png

PR #356, opened 2026-03-11
feat: Add Modern clock for desktop.

FaabFDB:
Took a while because i am busy with graduation stuff, but I can finally add the modern clock widget that i made to the repository.

It just adds a modern looking clock like the ones usually seen in wallpaper engine backgrounds.
You can change the font, colour and opacity.

The preview image is 960 wide but the height is a bit thinner as else it would just take up useless space. If it really needs to be 960x540 i can change it of course.

It has the free anurati regular font in the repository if that's alright. Just to make it easier for people to install it, if it's not allowed i'll remove it and tell people to download it online.

Also my first time doing something remotely with qml so any feedback is appreciated.


spiros132:
I would recommend using the desktopWidgetSettings so that if a user wants to, they can use multiple clocks on their desktop :)

FaabFDB:
@spiros132 Ahh i see what you mean, overlooked the part where if you change the settings it applies to all of them. Will do that thank you :)

FaabFDB:
I added the DesktopWidgetSettings functionality into the plugin. Hope it's alright like this, any recommendations are appreciated, or general code cleanup tips, i don't work with quickshell.

modern-clock/DesktopWidget.qml
```diff
@@ -0,0 +1,111 @@
...
+      : (widgetData?.dateColourPicker ?? "black")
+
+    readonly property color timeColour: colourChoice
+      ? Color.resolveColorKey(widgetData?.timeColourChoice ?? "secondary")
+      : (widgetData?.timeColourPicker ?? "black")
+
+  // Font
+  readonly property string customDayFont: widgetData?.customDayFont ?? Style.fontFamily ?? root.customFont ?? ""
+  readonly property string customDateFont: widgetData?.customDateFont ?? Style.fontFamily ?? root.customFont ?? ""
+  readonly property string customTimeFont: widgetData?.customTimeFont ?? Style.fontFamily ?? root.customFont ?? ""
```
spiros132: Is this really what you wanted? Since the `Style` will never be undefined / null because it's a singleton. That means that the `??` after the `Style` won't take effect, which means that the fonts will never be able to be the `root.customFont`

modern-clock/DesktopWidget.qml
```diff
@@ -0,0 +1,111 @@
...
+
+  readonly property real dayFontSize: widgetData?.dayFontSize !== undefined ? widgetData?.dayFontSize : 1.0
+  readonly property real dateFontSize: widgetData?.dateFontSize !== undefined ? widgetData?.dateFontSize : 1.0
+  readonly property real timeFontSize: widgetData?.timeFontSize !== undefined ? widgetData?.timeFontSize : 1.0
```
spiros132: If you prefer, this syntax does the exact same thing but a bit more concise `foo?.bar ?? "Bar"`. Meaning that if the `bar` variable in `foo` is also undefined then it will go to the string literal.

spiros132 (commented):
Looks good to me, just some minor things otherwise I'll merge this in :)

modern-clock/DesktopWidget.qml
FaabFDB: ahh i see, i will change this. Thank you!

modern-clock/DesktopWidget.qml
FaabFDB: yeah using 2 different methods now it seems. That's my bad, will change this to the method you suggested.

PR #527, opened 2026-03-30
Add Tamagotchi plugin

Xooter:
A lightweight virtual pet designed for passive interaction in the taskbar.

It’s intentionally low-maintenance, encouraging occasional check-ins rather than constant attention.


tamagotchi/Panel.qml
```diff
@@ -0,0 +1,65 @@
...
+        anchors.fill:    parent
+				anchors.margins: Style.spacingM
+				spacing:         30
+
+				StatBars {
+						Layout.fillWidth: true
+						hunger:      Tamagotchi.TamagotchiState.hunger
+						happiness:   Tamagotchi.TamagotchiState.happiness
+						cleanliness: Tamagotchi.TamagotchiState.cleanliness
+						energy:      Tamagotchi.TamagotchiState.energy
```
spiros132: For example, here you could pass either the `pluginApi` directly or the `pluginApi.mainInstance` as the "singleton".

spiros132 (commented):
Love the plugin idea!

Here is some feedback about the PR :)

spiros132 (commented):
Sorry for the late review, here is some more feedback about the PR :)

spiros132 (commented):
Two minor things otherwise I believe it's ready to get merged in :)

spiros132 (commented):
Looks great! Thank you so much for the PR! :D 

@ItsLemmy

PR #582, opened 2026-04-06
Update niri-auto-tile to v1.10.4

pir0c0pter0:
## Summary

Update niri-auto-tile from v1.4.2 to v1.10.4. This release reworks how the daemon handles window open and close events, reducing full-screen reprocessing and relying more on niri's native layout behavior. The result is noticeably smoother transitions, especially with many columns.

### Changes

- **Incremental open**: When a new window opens above the max-visible threshold, only the new column width is set. The daemon no longer walks every column from the first one, which used to cause the viewport to jump to the beginning of the workspace.

- **Full redistribute at threshold**: When column count reaches exactly max-visible (e.g. 3 to 4), a full redistribute runs to fill the screen. All other opens above the threshold are incremental.

- **Full redistribute below threshold**: When ONLY_AT_MAX is disabled, all columns are redistributed evenly from the first window — not just at the max. After redistribution, the viewport is pulled to the first column and centered to ensure all columns fill the screen.

- **Close delegates to niri**: Closing a window no longer triggers a full redistribute. The daemon lets niri handle the layout on its own.

- **Viewport fill on close**: When remaining columns still exceed max-visible after a close, the daemon waits briefly for niri to settle, then pulls off-screen columns into the viewport and centers them. Below max-visible, remaining columns are resized and re-centered.

- **Direct window ID focus**: All width operations now use focus-window --id instead of focus-column-first + focus-column-right walking. This prevents the viewport from scrolling to the beginning of the workspace.

- **Always redistribute by default**: Changed ONLY_AT_MAX default to false — columns are now redistributed from the first one, not only when reaching max-visible.

- **Logger compliance**: Replaced console.warn with Logger.w per Noctalia Plugin Guide.

- **Per-workspace independence preserved**: Each workspace and monitor continues to be handled independently.

Version bump: 1.4.2 -> 1.10.4

image


niri-auto-tile/PUBLISHING.md
spiros132, line 1: Is this file needed?

niri-auto-tile/README.md
```diff
@@ -2,25 +2,106 @@
...
+1. **Clone into the plugins directory:**
+
+   ```bash
+   git clone https://github.com/pir0c0pter0/niri-auto-tile.git \
+     ~/.config/noctalia/plugins/niri-auto-tile
+   ```
+
+2. **Enable** in Noctalia Settings > Plugins > niri-auto-tile
+
+3. **Add the bar widget** — drag "Auto-Tile" to your bar in Noctalia Settings > Bar
```
spiros132, line 75: This section isn't actually correct, just press on install and the plugin gets installed.

spiros132 (commented):
Some minor feedback about the PR :)

PR #586, opened 2026-04-06
feat: added translation support and panel mode

qholin:
- Added translation support and translation the plugin to Turkish.
- Added panel mode. It triggers when the bar widget's expanding direction is perpendicular to the bar.
- Added panel mode customization.
- Bug fix: Border color - none is now functional.
- Bug fix: Primary/secondary pill color - none is now functional.
- Updated README.md
- Bumped the version to 1.1.0


spiros132 (commented):
Some very minor feedback, otherwise it looks great! :D

special-workspaces/i18n/en.json
```diff
@@ -0,0 +1,97 @@
...
+        "label": "Focus Color",
+        "description": "Color of the drawer button when active."
+      },
+      "borderRadius": {
+        "label": "Button Radius",
+        "description": "Roundness of the drawer button (%1%)."
+      },
+      "size": {
+        "label": "Size",
+        "description": "Size of the drawer button (%1%)."
```
spiros132: Do not use this, prefer to use translation interpolations instead. For example:

#### Translation file
```json
{
  "example": "Foo: {bar}"
}
```

#### Qml file
```qml
// Will print "Foo: 10"
pluginApi?.tr("example", { bar: 10 })
```

spiros132 (commented):
Some small feedback about the PR :)

spiros132 (commented):
Thank you it looks great! :D 

@ItsLemmy

PR #595, opened 2026-04-08
Coloured CPU temps

simonjones49:
Added green, yellow and red for CPU temps


sys-info-widget/DesktopWidget.qml
```diff
@@ -201,7 +208,8 @@ DraggableDesktopWidget {
                     font.pointSize: Style.fontSizeL * widgetScale
                 }
                 NText {
-                    text: root.cpuUsage + " @ " + root.cpuTemp
+                    text: root.cpuUsage + " @ <font color='" + root.getTempHex(root.cpuTemp) + "'>" + root.cpuTemp + "</font>"
+                    textFormat: Text.RichText
```
spiros132: Instead of creating it like this, would it be possible to use the `color` that the `NText` component already has instead of doing it this way?

spiros132 (commented):
Looks great, thank you! :D 

@ItsLemmy

PR #597, opened 2026-04-08
Simple search filter for keybinds

nZo-sp:


spiros132 (commented):
Since @blackbartblues hasn't responded in a week and this is a pretty small PR, I'll merge it in. Everything looks good imo :D

PR #598, opened 2026-04-08
feat(monique): add Monique monitor profile switcher plugin

ToRvaLDz:
## Summary

- Adds the **Monique** plugin: switch monitor profiles from the Noctalia bar
- Shows the active profile as a tooltip on hover (optional label in the bar)
- Right-click opens a context menu to open Monique or access widget settings
- Left-click opens the profile switcher panel
- Supports 8 languages: en, de, es, fr, it, pt, ru, zh-CN

## Preview

preview

## Dependencies

Requires [`monique`](https://github.com/ToRvaLDz/monique) installed and available in `$PATH`.

## Test plan

- [ ] Plugin loads with `qs -c noctalia-shell`
- [ ] Bar widget shows icon and tooltip with active profile name
- [ ] Left-click opens panel with profile list
- [ ] Clicking a profile switches it and shows a toast notification
- [ ] Right-click → "Open Monique" launches the Monique GUI
- [ ] Right-click → "Widget settings" opens the settings page
- [ ] "Show profile name" toggle shows/hides the label
- [ ] Icon color setting applies correctly
- [ ] Settings persist across restarts
- [ ] Warning shown when `monique` is not installed


monique/Main.qml
```diff
@@ -0,0 +1,130 @@
...
+    command: ["monique", "--switch-profile", profileName]
+    stdout: StdioCollector {}
+    stderr: StdioCollector {}
+
+    onExited: function(exitCode) {
+      if (exitCode === 0) {
+        root.activeProfile = switchProcess.profileName
+        ToastService.showNotice(
+          pluginApi?.tr("toast.title"),
+          pluginApi?.tr("toast.switched") + " " + switchProcess.profileName,
```
spiros132: Would it be possible to use translation interpolations here instead of string concatenation?

monique/README.md
```diff
@@ -0,0 +1,28 @@
...
+- [Noctalia](https://github.com/arian-fallahpour/noctalia) ≥ 4.0.0
+- [Monique](https://github.com/ToRvaLDz/monique) installed and available in `$PATH`
+
+## Installation
+
+Copy or symlink the `monique/` folder into your Noctalia plugins directory:
+
+```bash
+ln -s ~/path/to/noctalia-monique/monique ~/.config/noctalia/plugins/monique
+```
```
spiros132: I'm not sure if this is needed since installation is by just pressing install.

spiros132 (commented):
Some very minor feedback! :)

monique/README.md
Cleboost: Needed only for developpment for hot reload but for most user it's not usefull

ToRvaLDz:
Thanks for the feedback!

- **Translation interpolation**: updated `toast.switched` across all i18n files to use `{profile}` as placeholder (e.g. `"Switched to {profile}"`) and replaced the string concatenation in `Main.qml` with `.replace("{profile}", ...)`, consistent with how other plugins in the repo handle it (e.g. `clipper`, `catwalk`).

- **Installation section**: removed from `README.md` — agreed, it's redundant since the plugin is installed via the registry. The manual symlink approach was only useful during local development.

monique/Main.qml
```diff
@@ -0,0 +1,130 @@
...
+    command: ["monique", "--switch-profile", profileName]
+    stdout: StdioCollector {}
+    stderr: StdioCollector {}
+
+    onExited: function(exitCode) {
+      if (exitCode === 0) {
+        root.activeProfile = switchProcess.profileName
+        ToastService.showNotice(
+          pluginApi?.tr("toast.title"),
+          pluginApi?.tr("toast.switched").replace("{profile}", switchProcess.profileName),
```
spiros132: So instead of doing it manually with the replace you can just use the interpolation feature that the `tr` function has. For example:

```qml
pluginApi?.tr("toast.switched", { profile: switchProcess.profileName })
```

ToRvaLDz: Updated to use tr() interpolation directly. 

Ty

spiros132 (commented):
Looks great!

@ItsLemmy

Cleboost (approved):

PR #619, opened 2026-04-09
Feature/coin flip settings

DanisDeveloper:
Add settings that allows change delay between animation frames


spiros132 (commented):
Looks good :D

PR #623, opened 2026-04-10
privacy-indicator: add camera filter regex

4hnme:
same as microphone filter, but for camera

### DIFF
--- a/privacy-indicator/Settings.qml
+++ b/privacy-indicator/Settings.qml
@@ -112,8 +112,8 @@ ColumnLayout {
 
     NTextInput {
       Layout.fillWidth: true
-      label: pluginApi?.tr("settings.camFilterRegex.label") || "Camera filter regex"
-      description: pluginApi?.tr("settings.camFilterRegex.desc") || "Regex pattern to filter out camera applications"
+      label: pluginApi?.tr("settings.camFilterRegex.label")
+      description: pluginApi?.tr("settings.camFilterRegex.desc")
       placeholderText: "droidcam"
       text: root.camFilterRegex
       onTextChanged: root.camFilterRegex = text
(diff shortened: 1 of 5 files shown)


privacy-indicator/Settings.qml
```diff
@@ -108,6 +109,15 @@ ColumnLayout {
       text: root.micFilterRegex
       onTextChanged: root.micFilterRegex = text
     }
+
+    NTextInput {
+      Layout.fillWidth: true
+      label: pluginApi?.tr("settings.camFilterRegex.label") || "Camera filter regex"
+      description: pluginApi?.tr("settings.camFilterRegex.desc") || "Regex pattern to filter out camera applications"
```
Cleboost:
```suggestion
      label: pluginApi?.tr("settings.camFilterRegex.label")
      description: pluginApi?.tr("settings.camFilterRegex.desc")
```

Cleboost (changes requested):

privacy-indicator/Settings.qml
4hnme: the same could probably be applied to `micFilterRegex`, since it is basically the same thing

Cleboost: if you have time you can remove all fallback for this plugin

Cleboost (approved):

spiros132:
@4hnme would you be able to change the panel so that the variable has an explicit type instead of an implicit one? So instead of using `property var`, use `property bool`, as shown in the code quality review.

spiros132 (commented):
Looks good, also can you bump the version as well so that people get the update? :D

privacy-indicator/manifest.json
Cleboost:
```suggestion
  "version": "1.2.9",
```

PR #634, opened 2026-04-11
Screen toolkit rework: UI, fixes, and improvements

ycf-anon:
This PR reworks the screen-toolkit plugin. It improves the UI, cleans up the structure, adds new tools, incorporates previous feedback, and fixes reported bugs. Overall, it should make the plugin more stable and consistent.

@spiros132 anytime :D

### DIFF
--- a/screen-toolkit/Annotate.qml
+++ b/screen-toolkit/Annotate.qml
@@ -485,7 +485,7 @@ Variants {
             height: overlayWin.currentStroke ? Math.abs(overlayWin.currentStroke.y2 - overlayWin.currentStroke.y1) : 0
             color:        "transparent"
             border.color: "#ffffff"
-            border.width: 1.5
+            border.width: Style.borderM
             opacity:      0.8
         }
         Rectangle {
@@ -791,11 +791,11 @@ Variants {
             radius:       Style.radiusL
             color:        Color.mSurface
             border.color: Style.capsuleBorderColor || "transparent"
-            border.width: Style.capsuleBorderWidth || 1
+            border.width: Style.capsuleBorderWidth || Style.borderS
             component ToolbarSeparator: Rectangle {
                 readonly property bool vertical: toolbar.useVertical
-                width:   vertical ? 28 : 1
-                height:  vertical ? 1 : 28
+                width:   vertical ? 28 : Style.borderS
+                height:  vertical ? Style.borderS : 28
                 color:   Color.mOnSurfaceVariant
                 opacity: 0.3
                 anchors.horizontalCenter: vertical ? parent.horizontalCenter : undefined
@@ -921,7 +921,13 @@ Variants {
                     visible: !parent.isVertical
                     Repeater {
                         model: 3
-                        Rectangle { width: 14; height: 2; radius: 1; color: Color.mOnSurfaceVariant; opacity: 0.6 }
+                        Rectangle {
+                            width:  Style.marginL
+                            height: Style.marginXXS
+                            radius: Style.radiusXXXS
+                            color:  Color.mOnSurfaceVariant
+                            opacity: 0.6
+                        }
                     }
                 }
                 Row {
@@ -930,7 +936,13 @@ Variants {
                     visible: parent.isVertical
                     Repeater {
                         model: 3
-                        Rectangle { width: 2; height: 14; radius: 1; color: Color.mOnSurfaceVariant; opacity: 0.6 }
+                        Rectangle {
+                            width:  Style.marginXXS
+                            height: Style.marginL
+                            radius: Style.radiusXXXS
+                            color:  Color.mOnSurfaceVariant
+                            opacity: 0.6
+                        }
                     }
                 }
                 MouseArea {
@@ -970,9 +982,9 @@ Variants {
                 "#44AAFF", "#CC44FF", "#FF44CC", "#FFFFFF", "#000000"
             ]
             readonly property var sizeDefs: [
-                { size: 2, label: root.mainInstance?.pluginApi?.tr("annotate.sizeS") ?? "S" },
-                { size: 4, label: root.mainInstance?.pluginApi?.tr("annotate.sizeM") ?? "M" },
-                { size: 7, label: root.mainInstance?.pluginApi?.tr("annotate.sizeL") ?? "L" }
+                { size: 2, label: root.mainInstance?.pluginApi?.tr("annotate.sizeS") },
+                { size: 4, label: root.mainInstance?.pluginApi?.tr("annotate.sizeM") },
+                { size: 7, label: root.mainInstance?.pluginApi?.tr("annotate.sizeL") }
             ]
             function doUndo() {
                 if (overlayWin.strokes.length > 0) {
@@ -1032,7 +1044,7 @@ Variants {
                         anchors.verticalCenter: parent.verticalCenter
                         color:        overlayWin.drawColor
                         border.color: overlayWin.showPopover ? Color.mPrimary : Qt.rgba(0, 0, 0, 0.2)
-                        border.width: overlayWin.showPopover ? 2 : 1
+                        border.width: overlayWin.showPopover ? Style.borderM : Style.borderS
                         scale: colorBtnH.containsMouse ? 1.1 : 1
                         Behavior on scale { NumberAnimation { duration: 80 } }
                         MouseArea {
@@ -1050,15 +1062,19 @@ Variants {
                     ActionBtn { iconName: "trash";          tip: "Clear all"; danger: true; onClicked: toolbar.doClear() }
                     SaveBtn {
                         iconName:  "copy"
-                        labelText: overlayWin.isSaving ? "Copying..." : "Copy"
-                        tip:       "Copy to clipboard"
+                        labelText: overlayWin.isSaving
+                            ? root.mainInstance?.pluginApi?.tr("annotate.copying")
+                            : root.mainInstance?.pluginApi?.tr("annotate.copy")
+                        tip:       root.mainInstance?.pluginApi?.tr("annotate.copyTip")
                         primary:   true
                         onClicked: overlayWin.flattenAndCopy()
                     }
                     SaveBtn {
                         iconName:  "device-floppy"
-                        labelText: overlayWin.isSaving ? "Saving..." : "Save"
-                        tip:       "Save to disk"
+                        labelText: overlayWin.isSaving
+                            ? root.mainInstance?.pluginApi?.tr("annotate.saving")
+                            : root.mainInstance?.pluginApi?.tr("annotate.save")
+                        tip:       root.mainInstance?.pluginApi?.tr("annotate.saveTip")
                         primary:   false
                         onClicked: overlayWin.flattenAndSave()
                     }
@@ -1101,7 +1117,7 @@ Variants {
                         anchors.horizontalCenter: parent.horizontalCenter
                         color:        overlayWin.drawColor
                         border.color: overlayWin.showPopover ? Color.mPrimary : Qt.rgba(0, 0, 0, 0.2)
-                        border.width: overlayWin.showPopover ? 2 : 1
+                        border.width: overlayWin.showPopover ? Style.borderM : Style.borderS
                         scale: colorBtnV.containsMouse ? 1.1 : 1
                         Behavior on scale { NumberAnimation { duration: 80 } }
                         MouseArea {
@@ -1117,8 +1133,8 @@ Variants {
                     ToolbarSeparator {}
                     ActionBtn { iconName: "corner-up-left"; tip: "Undo last stroke"; onClicked: toolbar.doUndo() }
                     ActionBtn { iconName: "trash";          tip: "Clear all"; danger: true; onClicked: toolbar.doClear() }
-                    ActionBtn { iconName: "copy";           tip: "Copy to clipboard"; onClicked: overlayWin.flattenAndCopy() }
-                    ActionBtn { iconName: "device-floppy";  tip: "Save to disk";      onClicked: overlayWin.flattenAndSave() }
+                    ActionBtn { iconName: "copy";           tip: root.mainInstance?.pluginApi?.tr("annotate.copyTip"); onClicked: overlayWin.flattenAndCopy() }
+                    ActionBtn { iconName: "device-floppy";  tip: root.mainInstance?.pluginApi?.tr("annotate.saveTip"); onClicked: overlayWin.flattenAndSave() }
                     ActionBtn { iconName: "x";              tip: "Close";             onClicked: toolbar.doClose() }
                     ToolbarSeparator {}
                     DragBtn { isVertical: true; anchors.horizontalCenter: parent.horizontalCenter }
@@ -1131,7 +1147,7 @@ Variants {
             radius:       Style.radiusL
             color:        Color.mSurface
             border.color: Style.capsuleBorderColor || "transparent"
-            border.width: Style.capsuleBorderWidth || 1
+            border.width: Style.capsuleBorderWidth || Style.borderS
             width:  toolbar.useVertical ? (popContent.implicitWidth  + Style.marginS) : (popContent.implicitWidth  + Style.marginM)
             height: toolbar.useVertical ? (popContent.implicitHeight + Style.marginM) : (popContent.implicitHeight + Style.marginS)
             x: toolbar.useVertical
@@ -1154,7 +1170,7 @@ Variants {
                         delegate: Rectangle {
                             width: 20; height: 20; radius: 10; color: modelData
                             border.color: overlayWin.drawColor === modelData ? Color.mPrimary : Qt.rgba(0, 0, 0, 0.15)
-                            border.width: overlayWin.drawColor === modelData ? 2 : 1
+                            border.width: overlayWin.drawColor === modelData ? Style.borderM : Style.borderS
                             scale: chH.containsMouse ? 1.2 : 1
                             Behavior on scale { NumberAnimation { duration: 80 } }
                             MouseArea {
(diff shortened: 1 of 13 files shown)


spiros132:
@ycf-anon I'll try to look at this PR when I have time but since it's a huge one it will take a bit of time for me. Would you be able to fix the auto code quality in the meantime? :)

ycf-anon:
@spiros132 Sure! One thing I wanted to check first: some of the flagged values look like they should stay hardcoded (e.g. icon shapes using Rectangles and size labels like "S/M/L"). What should I do with those?

spiros132:
> @spiros132 Sure! One thing I wanted to check first: some of the flagged values look like they should stay hardcoded (e.g. icon shapes using Rectangles and size labels like "S/M/L"). What should I do with those?

So for the flagged values that should be hard coded like icon shapes I believe you can forget those since that is a false positive that I should remove from the workflow. For the labels, tbh I'm not sure if the labels would change in different countries that have different alphabets, I think it might be a good idea to add them to the translations in that case. :)

ycf-anon:
Spiros132 hey,  I pushed fixes for the code quality stuff  
For the remaining review warnings, some of them kinda need to stay hardcoded otherwise it breaks the UI.

ycf-anon:
alright , some user say its lag the new region selector, did remove the animation, should fix the lag, was working fine for me , so i put this as draft until they confirm is good

ycf-anon:
ok, seems worked, so ready again

spiros132:
In the code quality review, a lot of those have to do with the `border.width: 1` or the `radius` field. When I have the time I'll try to make the workflow better so that it's easier to understand what it's complaining about. 

There is a border 1 that is called `Style.borderS` would you be able to use that one?

There are also `Style` fields for the radius, depending on if it's the radius for an "internal" component as in a button or the radius for an "external" component, aka a background or a rectangle. 

I would recommend looking through the following [file](https://github.com/noctalia-dev/noctalia-shell/blob/main/Commons/Style.qml) to check it out. Of course for the `radius: 14` specifically there isn't any style that is the exact that one, but you could use `Style.radiusS` or `Style.radiusM` which are 12 and 16 if you want, otherwise just leave it as it is. :)

ycf-anon:
Got it, thanks for the suggestion! I’ll test it on my side and see how it works. I actually tried this before, but for some reason it broke the UI, maybe I missed something.

For this PR, I’d prefer to keep things as they are since I’ve already made a lot of changes, and I’ll update everything to use the Style values in the next update🙏

screen-toolkit/Settings.qml
```diff
@@ -0,0 +1,211 @@
...
+        interval: 1000; repeat: true; running: true
+        onTriggered: root._previewNow = new Date().toString()
+    }
+
+    function _load() {
+        if (!pluginApi?.pluginSettings) return
+        _loaded = false
+        screenshotPath = pluginApi.pluginSettings.screenshotPath || ""
+        videoPath      = pluginApi.pluginSettings.videoPath      || ""
+        filenameFormat = pluginApi.pluginSettings.filenameFormat || ""
```
spiros132: I would suggest to skip the load functionality and directly bind the properties to the values. That way the values update automatically when changed instead of having to do it manually. For example:

```qml
property string example: pluginApi?.pluginSettings?.foo || "Foo"
```

screen-toolkit/Settings.qml
```diff
@@ -0,0 +1,211 @@
...
+                    anchors.verticalCenter: parent.verticalCenter
+                }
+                NText {
+                    text:        root.buildPreview(root.filenameFormat) + ".ext"
+                    pointSize:   Style.fontSizeXS
+                    color:       Color.mOnSurface
+                    font.family: "monospace"
+                    anchors.verticalCenter: parent.verticalCenter
+                    elide: Text.ElideRight
+                    width: parent.width - 32
```
spiros132: Is this 32 specific for something or is it for the margins? If it's for the margins I would suggest using something like `parent.width - (2 * Style.marginM)`

screen-toolkit/Record.qml
```diff
@@ -141,51 +151,50 @@ Item {
...
-                        "gifski --fps 20 --quality 100 -o " + root.gifPath +
-                        " " + framesDir + "/frame*.png 2>/dev/null && " +
-                        "rm -rf " + framesDir + " " + root.mp4Path
+                        "mkdir -p " + shellEscape(framesDir) + " && " +
+                        "ffmpeg -y -i " + shellEscape(root.mp4Path) +
+                        " -vf 'fps=24,scale=trunc(iw/2)*2:trunc(ih/2)*2:flags=lanczos'" +
+                        " " + shellEscape(framesDir) + "/frame%04d.png 2>/dev/null && " +
+                        "gifski --fps 24 --quality 97 -o " + shellEscape(root.gifPath) +
+                        " " + shellEscape(framesDir) + "/frame*.png 2>/dev/null && " +
+                        "rm -rf " + shellEscape(framesDir) + " " + shellEscape(root.mp4Path)
```
spiros132, line 197: For the next update, would you be able to remove these scripts from inline and create separate scripts? For example in a folder called "scripts" or something like that. It would make it much easier to review and also much more organized.

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
                     ToolbarSeparator {}
-
-                    ActionBtn { iconName: "corner-up-left"; tip: "Undo last stroke";  onClicked: toolbar.doUndo() }
+                    ActionBtn { iconName: "corner-up-left"; tip: "Undo last stroke"; onClicked: toolbar.doUndo() }
                     ActionBtn { iconName: "trash";          tip: "Clear all"; danger: true; onClicked: toolbar.doClear() }
-                    SaveBtn   { iconName: "copy";         labelText: overlayWin.isSaving ? "Copying..." : "Copy"; tip: "Copy to clipboard"; primary: true;  onClicked: overlayWin.flattenAndCopy() }
-                    SaveBtn   { iconName: "device-floppy"; labelText: overlayWin.isSaving ? "Saving..."  : "Save"; tip: "Save to ~/Pictures"; primary: false; onClicked: overlayWin.flattenAndSave() }
+                    SaveBtn {
+                        iconName:  "copy"
+                        labelText: overlayWin.isSaving ? "Copying..." : "Copy"
```
spiros132: Shouldn't this be in the translations?

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
+                        tip:       "Copy to clipboard"
+                        primary:   true
+                        onClicked: overlayWin.flattenAndCopy()
+                    }
+                    SaveBtn {
+                        iconName:  "device-floppy"
+                        labelText: overlayWin.isSaving ? "Saving..." : "Save"
+                        tip:       "Save to disk"
```
spiros132: All of these as well.

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
                         model: toolbar.toolDefs
                         ToolBtn { toolId: modelData.id; iconName: modelData.icon; tip: modelData.tooltip }
                     }
-
                     ToolbarSeparator {}
-
                     ZoomBtn {
-                        iconName: "zoom-out"; tip: "Zoom out"
+                        iconName:   "zoom-out"
+                        tip:        "Zoom out"
```
spiros132: Here as well.

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
-                    ActionBtn { iconName: "corner-up-left"; tip: "Undo last stroke";  onClicked: toolbar.doUndo() }
+                    ActionBtn { iconName: "corner-up-left"; tip: "Undo last stroke"; onClicked: toolbar.doUndo() }
                     ActionBtn { iconName: "trash";          tip: "Clear all"; danger: true; onClicked: toolbar.doClear() }
-                    ActionBtn { iconName: "copy";           tip: "Copy to clipboard";  onClicked: overlayWin.flattenAndCopy() }
-                    ActionBtn { iconName: "device-floppy";  tip: "Save to ~/Pictures"; onClicked: overlayWin.flattenAndSave() }
-                    ActionBtn { iconName: "x";              tip: "Close";              onClicked: toolbar.doClose() }
+                    ActionBtn { iconName: "copy";           tip: "Copy to clipboard"; onClicked: overlayWin.flattenAndCopy() }
+                    ActionBtn { iconName: "device-floppy";  tip: "Save to disk";      onClicked: overlayWin.flattenAndSave() }
+                    ActionBtn { iconName: "x";              tip: "Close";             onClicked: toolbar.doClose() }
+                    ToolbarSeparator {}
```
spiros132, line 1139: All of these hard-coded values as well.

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
                         "bash", "-c",
-                        "DEST=$([ -d \"$HOME/Pictures/Screenshots\" ] && echo \"$HOME/Pictures/Screenshots\" || echo \"$HOME/Pictures\"); " +
-                        "magick /tmp/screen-toolkit-annotate.png /tmp/screen-toolkit-overlay.png " +
-                        "-composite \"$DEST/" + filename + "\" && " +
-                        "rm -f /tmp/screen-toolkit-overlay.png && echo \"$DEST\""
+                        "if [ -d " + JSON.stringify(ssDir) + " ]; then DEST=" + JSON.stringify(ssDir) +
+                        "; elif [ -d " + JSON.stringify(picDir) + " ]; then DEST=" + JSON.stringify(picDir) +
+                        "; else exit 1; fi; " +
+                        "cp " + JSON.stringify(root.imagePath) + " \"$DEST/" + filename + "\" && " +
+                        "echo \"$DEST/" + filename + "\""
```
spiros132, line 1271: As before, in a future update, would it be possible to move all of this inline scripting into its own dedicated script?

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
+                    ]})
+                } else {
+                    var dest = custom + "/" + filename
+                    saveFileProc.exec({ command: [
+                        "bash", "-c",
+                        "cp " + JSON.stringify(root.imagePath) + " " + JSON.stringify(dest) + " && " +
+                        "echo " + JSON.stringify(dest)
```
spiros132, line 1278: Here as well if it's possible.

screen-toolkit/Annotate.qml
```diff
@@ -525,374 +763,532 @@ Variants {
...
+                            "echo \"$DEST/" + filename + "\""
+                        ]})
+                    } else {
+                        var dest2 = custom + "/" + filename
+                        saveFileProc.exec({ command: [
+                            "bash", "-c",
+                            "magick /tmp/screen-toolkit-annotate.png /tmp/screen-toolkit-overlay.png " +
+                            "-composite " + JSON.stringify(dest2) + " && " +
+                            "rm -f /tmp/screen-toolkit-overlay.png && " +
+                            "echo " + JSON.stringify(dest2)
```
spiros132, line 1305: And all of these as well.

screen-toolkit/Main.qml
```diff
@@ -7,91 +7,125 @@ import qs.Services.UI
...
+        pluginApi.pluginSettings.stateIsRunning     = root.isRunning
+        pluginApi.pluginSettings.stateActiveTool    = root.activeTool
+        pluginApi.pluginSettings.stateMirrorVisible = mirrorOverlay?.isVisible ?? false
         pluginApi.saveSettings()
     }
 
-    // ── Settings shortcuts ────────────────────────────────────
     readonly property string selectedOcrLang: pluginApi?.pluginSettings?.selectedOcrLang || "eng"
+    readonly property bool   isNiri:          (Quickshell.env("NIRI_SOCKET")                  || "") !== ""
+    readonly property bool   isHyprland:      (Quickshell.env("HYPRLAND_INSTANCE_SIGNATURE")  || "") !== ""
```
spiros132: Instead of this, I would suggest using the noctalia compositor service to check if we are in niri or hyprland.

screen-toolkit/Main.qml
```diff
@@ -332,103 +421,86 @@ Item {
...
-                    + "; grim -g \"${GX},${GY} 11x11\" \"" + file + "\" 2>/dev/null || exit 1"
-                    + "; HEX=$(magick \"" + file + "\" -alpha off -format '#%[hex:p{5,5}]' info:- 2>/dev/null)"
-                    + "; [ -n \"$HEX\" ] && printf '%s|%s' \"$HEX\" \"" + file + "\" || exit 1"
+            var cmd = "COORDS=$(slurp -p 2>/dev/null) || exit 1; " +
+                      "X=${COORDS%%,*}; REST=${COORDS#*,}; Y=${REST%% *}; " +
+                      "GX=$((X > 5 ? X - 5 : 0)); GY=$((Y > 5 ? Y - 5 : 0)); " +
+                      "grim -g \"${GX},${GY} 11x11\" " + shellEscape(file) + " 2>/dev/null || exit 1; " +
+                      "magick " + shellEscape(file) + " -alpha off " +
+                      "-format '%[fx:int(255*u.p{5,5}.r)] %[fx:int(255*u.p{5,5}.g)] %[fx:int(255*u.p{5,5}.b)]' " +
+                      "info:- 2>/dev/null"
```
spiros132, line 446: As before, could this become a script by itself in a future update?

screen-toolkit/Main.qml
```diff
@@ -332,103 +421,86 @@ Item {
...
+                "fi && " +
+                "TEXT=$(tesseract " + shellEscape(tmp) + " stdout -l " + shellEscape(lang) + " --psm " + psm + " --oem 1 2>/dev/null) && " +
+                "if [ $(printf '%s' \"$TEXT\" | tr -d '[:space:]' | wc -c) -lt 4 ]; then " +
+                "  TEXT2=$(magick " + shellEscape(tmp) + " -threshold 85% stdout | " +
+                "          tesseract - stdout -l " + shellEscape(lang) + " --psm " + psm + " --oem 1 2>/dev/null); " +
+                "  if [ $(printf '%s' \"$TEXT2\" | tr -d '[:space:]' | wc -c) -gt $(printf '%s' \"$TEXT\" | tr -d '[:space:]' | wc -c) ]; then " +
+                "    TEXT=\"$TEXT2\"; fi; " +
+                "fi && " +
+                "printf '%s' \"$TEXT\"; rm -f " + shellEscape(tmp)
+
```
spiros132: All of this as well if it's possible.

screen-toolkit/Main.qml
```diff
@@ -438,68 +510,113 @@ Item {
...
+        var file = "/tmp/screen-toolkit-lens.png"
+        var cmd =
+            _grimRegionCmd(file) + " || { notify-send -u critical 'Screen Toolkit' 'Capture failed'; exit 1; }; " +
+            "notify-send 'Screen Toolkit' 'Uploading to Lens...' 2>/dev/null; " +
+            "RESP=$(curl -sS --connect-timeout 5 --max-time 30 -F 'files[]=@" + file + "' 'https://uguu.se/upload' 2>/dev/null); " +
+            "URL=$(echo \"$RESP\" | jq -r '.files[0].url // empty' 2>/dev/null); " +
+            "rm -f " + file + "; " +
+            "if [ -n \"$URL\" ] && [ \"$URL\" != \"null\" ]; then " +
+            "  xdg-open \"https://lens.google.com/uploadbyurl?url=$URL\" >/dev/null 2>&1; exit 0; " +
+            "else notify-send -u critical 'Screen Toolkit' 'Upload failed or timed out'; exit 1; fi"
```
spiros132: Here as well.

screen-toolkit/Main.qml
```diff
@@ -438,68 +510,113 @@ Item {
...
+    Timer {
+        id: launchPinFile
+        interval: 200; repeat: false
+        onTriggered: {
+            pinFileProc.exec({ command: [
+                "bash", "-c",
+                "if command -v zenity >/dev/null 2>&1; then " +
+                "zenity --file-selection --title='Pin image' --file-filter='Images | *.png *.jpg *.jpeg *.webp *.gif *.bmp' 2>/dev/null; " +
+                "elif command -v kdialog >/dev/null 2>&1; then " +
+                "kdialog --getopenfilename '' 'Images (*.png *.jpg *.jpeg *.webp *.gif *.bmp)' 2>/dev/null; fi"
```
spiros132, line 576: I won't write it on every single one but I hope you get my point xD

spiros132 (commented):
Some feedback about the PR :)

ycf-anon:
Thanks for the fast review man, really appreciate you taking the time. I’ll check everything you mentioned and fix it

screen-toolkit/Record.qml
ycf-anon: could you please show me how you’d handle moving those scripts out of inline?

spiros132: Of course (excuse my formatting since I'm on the phone). So basically I would suggest understanding which variables that are needed for these scripts first, let's say you need variable `a` and variable `b`. Then you create a script, I would suggest most probably bash since that's what the scripts are in, which looks something like this:

```bash
#!/bin/bash

# The first argument to this script we define as it should always be variable `a`
a=$0

# The second argument is always `b`
b=$1


# Maybe double check first here that both variables look good
if [ -z "$a" ] || [ -z "$b" ]; then
echo "This script is supposed to be used like this: script argument_a argument_b"
exit 1
fi

# Then do all the actions you need
mkdir "$a"
touch "$b"

echo "etc etc..."

exit 0
```

Then in the qml code where you have the process the command itself might look like this:

```qml
command: [ "/path/to/script", "argument_a", "argument_b" ]
```

If you have any other questions, don't hesitate to ask! :D

ycf-anon: Got it, i’ll try it, if something breaks, i’ll let you know  , thank you

spiros132 (commented):
Looks great! Great job on this huge PR :D

spiros132:
@ycf-anon are you done with everything? Should I merge it in?

ycf-anon:
Yup, all good here, thanks for the review and help

PR #636, opened 2026-04-12
Fix stuck loading state for failed VPN actions

Tr0sT:
This fixes a bug in the network-manager-vpn plugin where a failed connect/disconnect attempt leaves the item stuck in a loading
  state.

  Changes:
  - clear pending/loading state when `nmcli connection up/down` exits with an error
  - keep the VPN action button usable for retry after a failed attempt
  - expose `isLoading` on the main instance for the bar widget

### DIFF
--- a/network-manager-vpn/Main.qml
+++ b/network-manager-vpn/Main.qml
@@ -14,6 +14,7 @@ QtObject {
 
     property var vpnList: []
     property real connectedCount: 0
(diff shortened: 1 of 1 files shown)


spiros132 (commented):
@nZo-sp

nZo-sp:
Seems ok,thank you. @spiros132

PR #637, opened 2026-04-12
i18n (video-wallpaper): updated german translations

D3rJust1n:
updated german translations for the `video-wallpaper` plugin
also added a missing translation string in `i18n/en.json`


spiros132 (commented):
Thank you so much for the translation! :D 
@ItsLemmy

PR #639, opened 2026-04-13
feat(ip-monitor): change IP fetch API from ipinfo.io to ip-api.com

neyfua:
@deep4lpha I changed the IP fetch API in the plugin to support when you connect to Warp or VPN, the one were used isn't fetched when connecting to Warp or VPN because of the rates limit.

I also removed the IPC commands from the Panel.qml and added it into Settings.qml instead, I think users would prefer seeing it inside Settings instead of Panel. Also little adjustment on the IP details since the new API doesn't provide some of the old datas.

Hope this new one helps with the Warp and VPN connection failed to fetch IP issue when using ipinfo.io.

If you have time please review these and tell me if there's anything that should be changed, really thanks for this cool plugin :).

### DIFF
--- a/ip-monitor/BarWidget.qml
+++ b/ip-monitor/BarWidget.qml
@@ -1,12 +1,12 @@
 import QtQuick
 import QtQuick.Layouts
 import Quickshell
+import "." as Local
 import qs.Commons
 import qs.Modules.Bar.Extras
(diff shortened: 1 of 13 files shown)


neyfua:
@deep4lpha if you have some time pls review the changes and let me know if there's anything else that needed fix/change :)

PR #640, opened 2026-04-13
replacement (osk-toggle): replaces squeekboard-toggle, works also with wvkbd

blackteaextract:
This replaces the existing `squeekboard-toggle` plugin with a consolidated
`osk-toggle` that handles both Squeekboard and wvkbd from a single widget.
The Squeekboard behavior is unchanged — it's purely additive.

## What's new

**wvkbd support**
wvkbd is now a fully supported backend. The plugin owns the wvkbd process
(kill-on-start, relaunch with `--hidden`), toggles via SIGUSR1/SIGUSR2, and
tracks visibility across unexpected exits.

**Auto-detection**
The default backend is `"auto"`: at startup the plugin does a one-shot D-Bus
check for `sm.puri.OSK0`. If it's owned, Squeekboard is used; otherwise wvkbd.
Users can pin a backend explicitly in settings.

**Backend selector in settings**
The backend setting is a dropdown (Auto / Squeekboard / wvkbd).

**Informative tooltip when unavailable**
When no backend is loaded, the widget shows a localized message pointing the
user to settings.

## Breaking change

The plugin ID changes from `squeekboard-toggle` to `osk-toggle`. Existing
users will need to re-add the widget — settings won't migrate automatically.

## Files

```
osk-toggle/
├── Main.qml               — backend loader + auto-detection
├── SqueekboardBackend.qml — squeekboard logic (extracted from old plugin)
├── WvkbdBackend.qml       — wvkbd logic (new)
├── BarWidget.qml
├── Settings.qml           — backend dropdown, same other settings
├── manifest.json          — new id/version, same structure
└── i18n/                  — 16 locales, all updated
```


spiros132:
Since removing the plugin completely might be weird for some users, I would suggest to instead leave it there, but add a deprecated section in the README. :)

blackteaextract:
> Since removing the plugin completely might be weird for some users, I would suggest to instead leave it there, but add a deprecated section in the README. :)

Yes we can certainly do that! :)
Should I make a new PR?

spiros132:
> > Since removing the plugin completely might be weird for some users, I would suggest to instead leave it there, but add a deprecated section in the README. :)
> 
> Yes we can certainly do that! :) Should I make a new PR?

It's completely up to you if you want to create a new PR. I usually would suggest having one PR for one plugin, but if you're only going to change a few lines for the `squeekboard-toggle` plugin, you don't need to do that.

blackteaextract:
> > > Since removing the plugin completely might be weird for some users, I would suggest to instead leave it there, but add a deprecated section in the README. :)
> > 
> > 
> > Yes we can certainly do that! :) Should I make a new PR?
> 
> It's completely up to you if you want to create a new PR. I usually would suggest having one PR for one plugin, but if you're only going to change a few lines for the `squeekboard-toggle` plugin, you don't need to do that.

Thanks!
Then I will open a new PR, otherwise it will get messy... :)

PR #641, opened 2026-04-13
add (not-just-text): plugin that displays a short message in the bar

blackteaextract:
## Summary

Adds the **Not Just Text** bar widget plugin (`not-just-text`).

The widget displays a short message in the bar with three mutually exclusive modes:

- **Just text** — static string the user types directly (default: `Hello`)
- **Fortune mode** — random quote via `fortune -s`, filtered to single-line entries ≤ 60 chars, retries up to 10 times
- **List mode** — random line from a user-provided text file; falls back to the bundled `examples.txt`

Both fortune and list modes refresh automatically on wallpaper change (debounced 300 ms to handle per-screen events).

preview

## Files

| File | Purpose |
|---|---|
| `Main.qml` | Shared state: runs `fortune` process and reads text file; exposes `fortuneText` / `listText` |
| `BarWidget.qml` |  |
| `Settings.qml` | Settings UI with mode toggles and conditional inputs |
| `manifest.json` | Plugin metadata, entry points, default settings |
| `examples.txt` | Bundled default list file (kaomoji + short phrases) |
| `i18n/*.json` | Translations for 17 locales |

## Behaviour notes

- Lines starting with `# ` (hash + space) and blank lines are skipped in list mode; bare `#tag` entries are valid
- Fortune: exits with `⚠ fortune not installed` message and logs an error if the binary is missing (exit code 127)
- List: logs a warning and sets `listText = ""` if the file is unreadable
- Fortune and list modes are mutually exclusive — enabling one in Settings.qml disables the other


not-just-text/Main.qml
```diff
@@ -0,0 +1,143 @@
...
+        onTriggered: {
+            if (root.fortuneEnabled) triggerFortune();
+            if (root.listEnabled) pickFromFile();
+        }
+    }
+
+    Connections {
+        target: WallpaperService
+        function onWallpaperChanged(screenName, path) {
+            if (root.fortuneEnabled || root.listEnabled) debounce.restart();
```
spiros132, line 74: Would be nice if there was a toggle as well for this, since some people might not want their text to change when the wallpaper changes.

not-just-text/Main.qml
```diff
@@ -0,0 +1,143 @@
...
+                    Logger.w("NotJustText", "Gave up after", root._maxRetries, "retries finding a short single-line fortune");
+                    root.fortuneText = "(╯°□°）╯︵ ┻━┻";
+                    root._retries = 0;
+                }
+            }
+        }
+        onExited: (exitCode, exitStatus) => {
+            if (exitCode === 127) {
+                Logger.e("NotJustText", "fortune is not installed — install it to use fortune mode");
+                root.fortuneText = "⚠ fortune not installed";
```
spiros132: Shouldn't this also be included in the translations?

not-just-text/settings.json
spiros132, line 1: This file shouldn't be included in the PR

spiros132 (commented):
Looks good, just some minor feedback about the PR! :D

not-just-text/Main.qml
blackteaextract: i thought of it as well, but then they would just set the text via the "just text" mode, right?
But sure, i can implement it :)

not-just-text/Main.qml
blackteaextract: yes, will do!

not-just-text/Main.qml
spiros132: Oh yeah you are right actually, since then it would mean that the users would need a way to select the fortune or the a text from the list.

blackteaextract: There is an option to disable the change now, if someone wants to have one fortune per session. Also I added the option to adjust the length of the fortune, since display space may vary.

spiros132 (commented):

blackteaextract:
> Looks great, thank you for the PR! :D

Thank you for your review! :D

PR #642, opened 2026-04-13
initial release khal-agenda

simonjones49:
I have left in the kitty line
Maybe I need to simplify this 
 command: ["kitty", "--config",  "~/.config/kitty/calendar.conf", "--class",  "khal", "-e", "ikhal"]


khal-agenda-widget/i18n/en.json
```diff
@@ -0,0 +1,8 @@
+{
+  "widget": {
+    "loading": "Loading...",
+    "no_events": "No upcoming events",
+    "heading": "Agenda for 7 days.",
+    "button":"📅"
```
spiros132: I would suggest to use an icon that already exists in noctalia instead of this. That way it would be more consistent with the rest of the shell.

khal-agenda-widget/DesktopWidget.qml
```diff
@@ -0,0 +1,96 @@
...
+                Layout.fillWidth: true
+
+                NText {
+                    text: pluginApi.tr("widget.heading")
+                    color: Color.mOnSurfaceVariant
+                    font.pointSize: Style.fontSizeL * widgetScale
+                    Layout.fillWidth: true
+                }
+
+                Button {
```
spiros132: Prefer to use noctalias native widget instead of qt's. That way it is more consistent with the rest of the shell. For example noctalia has `NButton` instead of `Button`.

khal-agenda-widget/DesktopWidget.qml
```diff
@@ -0,0 +1,96 @@
...
+                    flat: true
+                    background: null
+                    onClicked: {
+                        openIkhal.running = false
+                        openIkhal.running = true
+                    }
+                }
+            }
+
+            ScrollView {
```
spiros132: Here as well, noctalia has `NScrollView`, prefer to use that one instead of the regular `ScrollView`.

khal-agenda-widget/DesktopWidget.qml
```diff
@@ -0,0 +1,96 @@
...
+        running: root.pluginApi !== null
+        stdout: StdioCollector {
+            onTextChanged: root.calendarVal = text.trim().replace(/\n\s*\n/g, '\n')
+        }
+    }
+
+    Process {
+        id: openIkhal
+        // command: ["kitty", "--config",  "~/.config/kitty/calendar.conf", "--class",  "khal", "-e", "ikhal"]
+        command: ["kitty",  "--class",  "khal", "-e", "ikhal"]
```
spiros132: This wouldn't work with people that don't have kitty. I believe there is a command to try to open a command with the default terminal, but I've forgotten what it is. Otherwise, like some of the other plugins, you could try to find which terminal is installed, and launch that one.

spiros132 (commented):
It looks good, just some minor feedback! :D

spiros132 (commented):

PR #645, opened 2026-04-13
fix (not-just-text): fix missing Settings

blackteaextract:
I am so sorry @spiros132 , I have uploaded the wrong/ old Settings.qml.
The improved Settings regarding the wallpaper change are now displayed and working.


spiros132 (commented):
Looks good :)

PR #646, opened 2026-04-13
feat: Add remix of screenshot-record

Mathew-D:
Adds sway support with window detection
Adds mangoWC support with just area selection
Add setting of where to save video and images
Still No Niri support
Common Objects that gets extended for window detection in Sway and Hyrland


spiros132:
This PR adds a completely new plugin, but why not do the updates on the already existing screenshot-and-record?

PR #653, opened 2026-04-14
Improve slowbongo

tmarti2:
This PR was originaly on the slowbongo repository but was never merged. Since slowbongo is now part of Noctalia I'm doing it again.

This MR:
- Add Enter as a big hit key
- Reorganize settings a bit (I think it is better this way but debatable)
- Add default values / reset to sliders
- Use NColorChoice widget. When I first saw your plugin I thought it was a better way to chose colors than comboboxes used by Noctalia, so I added a new widget for them. It is simpler than yours but it looks good and it simplifies your code so I went for it
- Fix a bug where slowbongo would check indefinitely for devices instead of stop after 3 fails 
- And the big part: I wanted a way for bongo cat to wait for key release to lift its paws. This is a minimal implementation:
  - I did not touch the music part, which uses the same function than before
  - The behavior is good for keyboard/mouse. I had to block BTN_TOUCH_* to avoid having 2 signals with touch gestures
  - A more accurate way to handle this would be to have counters for the number of (big) key pressed, but this complicates the code a lot so I went for the easy solution.

Thanks

### DIFF
--- a/slowbongo/Main.qml
+++ b/slowbongo/Main.qml
@@ -292,9 +292,9 @@ Item {
                 stdout: SplitParser {
                     onRead: data => {
                         if (data.includes("EV_KEY") && data.includes("value 1")) {
-                            // Detect spacebar for double slap (both paws)
-                            const isSpace = data.includes("KEY_SPACE");
-                            root.onKeyPress(isSpace);
+                            // Detect spacebar/enter for double slap (both paws)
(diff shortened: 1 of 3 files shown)


spiros132:
Would you be able to fix all the automatic code review? :)

tmarti2:
> Would you be able to fix all the automatic code review? :)

Well, this have nothing to do (except one maybe) with this MR, but I can do it

spiros132 (commented):
It looks good, thank you for the PR! :D

PR #654, opened 2026-04-14
refactor(ip-monitor): change IP fetch API from ipinfo.io to ip-api.com

neyfua:


neyfua:
@deep4lpha sorry for the mention but pls review this instead :), thanks

deep4lpha:
Looks good - what is the motivation behind changing backend? Perhaps we could make a configurable backend for ip checks?

neyfua:
> Looks good - what is the motivation behind changing backend? Perhaps we could make a configurable backend for ip checks?

The reason was because when you connect to Cloudflare WARP or VPN it'll fail to fetch the IP because of the rates limit, with this new one it's still success even if you connect to Cloudflare WARP or VPN

neyfua:
@deep4lpha do u think this is good enough to merge?

spiros132:
@deep4lpha I'll merge this in since you said before that it looks good

PR #660, opened 2026-04-16
feat(clipboard): add clipboard history panel

yanekyuk:
## Summary

- Fix NGridView image grid reverting to single column after switching from GridView to NGridView — use `availableWidth` instead of `width` to account for scrollbar reservation
- Fix buttons and timestamp flickering when hovering over action buttons — use `HoverHandler` (geometry-based) instead of `rowArea.containsMouse` (z-order-dependent) for stable hover state
- Move runtime files to cache dir (`~/.cache/noctalia/plugins/clipboard/`) instead of config dir
- Drop unverified `Hyprland` and `Niri` tags — plugin works on any Wayland compositor

## Changes

- `Panel.qml`: `cellWidth: Math.floor(availableWidth / 2)` — NGridView exposes `availableWidth` as the inner GridView's usable width (minus scrollbar)
- `ClipboardItem.qml`: Add `HoverHandler { id: cardHover }` on the card root; replace all `rowArea.containsMouse` references with `cardHover.hovered`
- `Main.qml`: Use cache dir for all runtime files; template literals for numeric-safe shell commands; input validation on `copyEntry(id)`
- `manifest.json`: Remove `Hyprland` and `Niri` tags (compositor-agnostic — works on any Wayland compositor)


yanekyuk:
> ### Automatic Manifest Check
> ### File: clipboard/manifest.json
> 
>     * [(H)](660) Line 7: The repository field needs to point to the "https://github.com/noctalia-dev/noctalia-plugins" github repository!
> 
> 
> ```json
> +  "repository": "https://github.com/yanekyuk/clipboard",
> ```

Fixed

clipboard/manifest.json
```diff
@@ -0,0 +1,27 @@
+{
+  "id": "clipboard",
+  "name": "Clipboard",
+  "version": "1.0.1",
```
Cleboost: why 1.0.1 and not 1.0.0?

clipboard/Main.qml
```diff
@@ -0,0 +1,1198 @@
...
+            // x-special as plain text.
+            //
+            // The URI is passed via $1 (not interpolated) so shell
+            // metacharacters in the path cannot execute. printf's
+            // format string contains only constants so "copy\n" and
+            // the trailing newline are emitted verbatim.
+            copyProc.command = ["bash", "-c",
+                'printf "copy\n%s\n" "$1" | wl-copy --type x-special/gnome-copied-files', "--", uri];
+        } else {
+            copyProc.command = ["bash", "-c", "cliphist decode " + id + " | wl-copy"];
```
Cleboost:
```suggestion
            copyProc.command = ["bash", "-c", 'cliphist decode "$1" | wl-copy', "--", sid];
```

clipboard/Main.qml
```diff
@@ -0,0 +1,1198 @@
...
+    function copy(id) {
+        if (!id)
+            return;
+        // File entries need x-special/gnome-copied-files so paste targets
+        // (Nautilus and most GTK file managers) recognise them as file
+        // references rather than plain text. See the branch below for
+        // the full rationale. Look up the item type; if it's a file,
+        // build the file:// URI and pass it via $1 (not interpolated)
+        // so shell metacharacters in the path cannot execute.
+        const sid = String(id);
```
Cleboost, line 653:
```suggestion
        const sid = String(id);
        if (!/^[0-9]+$/.test(sid))
            return;
```

like `getImage(id)` and `getFileMeta(id)`

clipboard/manifest.json
yanekyuk: Because I implemented the changes suggested by Automatic Code Quality Review.

clipboard/Main.qml
yanekyuk: thanks

clipboard/Main.qml
yanekyuk: thanks

clipboard/i18n/en.json
spiros132, line 1: English is the only translation that has 47 lines instead of 28 like the rest of the translations have. Do check out if you can update the rest as well.

clipboard/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```bash
+git clone https://github.com/yanekyuk/clipboard ~/.config/noctalia/plugins/clipboard
+```
+
+Then reload the shell:
+
+```bash
+qs kill -c noctalia-shell; sleep 1; qs -d -c noctalia-shell
+```
```
spiros132, line 35: This installation section isn't correct.

clipboard/ClipboardItem.qml
```diff
@@ -0,0 +1,661 @@
...
+
+    // --- Hover / press detection --------------------------------------------
+
+    // Hover state drives:
+    //   - card border accent (idle: mOutline; hover/press: mPrimary)
+    //   - visibility of the delete button (hidden until hover)
+    //   - pressed fill overlay (mPrimary @ 12% opacity while held)
+    HoverHandler {
+        id: hoverHandler
+    }
```
spiros132: Would it be possible to use the `MouseArea` for the hover control instead of a seperate `HoverHandler`?

clipboard/ClipboardItem.qml
```diff
@@ -0,0 +1,661 @@
...
+//   - Hover: reveal an icon-only delete button that removes the entry and
+//     fires a separate toast.
+//
+// Contracts followed:
+//   - docs/specs/plugin-qml-idioms.md (component composition — internal
+//     components receive pluginApi from the parent; null-guard everything;
+//     ToastService usage for transient confirmations; pointSize for NText).
+//   - Main.qml exposes copy(id) / remove(id) via pluginApi.mainInstance.
+Item {
+    id: card
```
spiros132, line 31: As a general tip, usually in qml the root object is called `root` instead of a separate name. You don't need to change this if you don't want to though.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    property int pinnedRevision: 0
+
+    // Absolute path to the pinned JSON file. Resolved at construction so
+    // both pinnedFile.path and the save function below read the same value.
+    // Location mirrors the canonical Clipper example in
+    // docs/specs/plugin-qml-idioms.md §FileView pattern — co-locating with
+    // settings.json under ~/.config/noctalia/plugins/<id>/ keeps all
+    // plugin-owned state under the same directory the shell already
+    // manages.
+    readonly property string pinnedPath: (Quickshell.env("HOME") || "") + "/.config/noctalia/plugins/clipboard/pinned.json"
```
spiros132: Do not use the path to the config. For per use cases prefer to use the cache path that noctalia has. For example: `~/.cache/noctalia/plugins/clipboard/...`

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    // settings.json via pluginApi.saveSettings(), so the path is already
+    // guaranteed to exist before copiedAtWriteProc fires.
+    //
+    // `copiedAtLoaded` is the gate that suppresses the first write: until
+    // onLoaded assigns the parsed map into root.copiedAt, listProc.onExited
+    // must not call saveCopiedAt() — otherwise the first refresh after
+    // startup can race the FileView read and erase persisted timestamps.
+    FileView {
+        id: copiedAtFile
+        path: Quickshell.env("HOME") + "/.config/noctalia/plugins/clipboard/copied-at.json"
```
spiros132: Here as well, prefer to use the cache path instead of the config path.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+                    if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
+                        root.copiedAt = parsed;
+                    }
+                }
+            } catch (e) {
+                // Malformed JSON on disk — keep root.copiedAt at its default
+                // ({}) so listProc.onExited can repopulate from cliphist.
+                // No Logger call: a first-run user has no file yet and the
+                // adapter's printErrors:false already silences the FileView
+                // read itself.
```
spiros132: Here might be a good idea to add a log with the `Logger.e("Clipboard Plugin", "...")` method.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    // Process is already running, we skip this tick — the next refresh
+    // will write a fresh snapshot, and the map mutates only on
+    // listProc.onExited so no intermediate state can be lost.
+    function saveCopiedAt() {
+        if (!root.copiedAtLoaded)
+            return;
+        if (copiedAtWriteProc.running)
+            return;
+        const payload = JSON.stringify(root.copiedAt || {});
+        const target = Quickshell.env("HOME") + "/.config/noctalia/plugins/clipboard/copied-at.json";
```
spiros132: As before, prefer to use the cache. Also since you're referring to this path multiple times, might be a good idea to make a `readonly property` for this variable.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+        // `printf %s` (no trailing newline in the format string) emits the
+        // JSON verbatim. Passing the payload and path as $1 / $2 keeps
+        // them out of the shell's interpolation path — payload contains
+        // arbitrary JSON and the path contains $HOME, both of which are
+        // safe as argv elements but dangerous if embedded in the command
+        // string. The same argv-as-positional policy is used by copy()
+        // and metaProc — see those for the rationale.
+        copiedAtWriteProc.command = ["bash", "-c",
+            'printf %s "$1" > "$2"', "--", payload, target];
```
spiros132, line 314: You can just add the variables directly in the string itself by using the `` characters. For example: 
```bash
`printf %s "${payload}" > "${target}"`
```

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+            // x-special as plain text.
+            //
+            // The URI is passed via $1 (not interpolated) so shell
+            // metacharacters in the path cannot execute. printf's
+            // format string contains only constants so "copy\n" and
+            // the trailing newline are emitted verbatim.
+            copyProc.command = ["bash", "-c",
+                'printf "copy\n%s\n" "$1" | wl-copy --type x-special/gnome-copied-files', "--", uri];
+        } else {
+            copyProc.command = ["bash", "-c", 'cliphist decode "$1" | wl-copy', "--", sid];
```
spiros132: Here as well you can use the `` to put the variables inside the string.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    // numeric id as `$1` (not interpolated into the shell string) so shell
+    // metacharacters in the id — implausible, but the same policy as
+    // copy() / metaProc — cannot execute. grep -P for the literal tab and
+    // -m1 so only the first matching line is forwarded.
+    function remove(id) {
+        if (!id)
+            return;
+        const sid = String(id);
+        removeProc.command = ["bash", "-c",
+            'cliphist list | grep -Pm1 "^$1\\t" | cliphist delete', "--", sid];
```
spiros132: Here as well.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+        const entry = list[index];
+        const preview = String(entry.preview || "");
+        const type = String(entry.type || "text");
+        if (type === "file") {
+            // File-URI fast path: same MIME treatment as copy(). `$1` is
+            // not interpolated into the shell string so metacharacters in
+            // the path cannot execute.
+            const uri = preview.indexOf("file://") === 0 ? preview : "file://" + preview;
+            copyProc.command = ["bash", "-c",
+                'printf "copy\n%s\n" "$1" | wl-copy --type x-special/gnome-copied-files', "--", uri];
```
spiros132, line 867: Here as well.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    // Drain one item from decodeQueue into decodeProc. No-op if the queue is
+    // empty or the process is already running. Called by getImage() on a
+    // cache miss and by decodeProc.onExited to chain the next decode.
+    function _processNextDecode() {
+        if (decodeProc.running || root.decodeQueue.length === 0)
+            return;
+        const sid = root.decodeQueue[0];
+        root.decodeQueue = root.decodeQueue.slice(1);
+        root.decodingId = sid;
+        decodeProc.command = ["bash", "-c", "cliphist decode " + sid + " > /tmp/clipboard-" + sid + ".png"];
```
spiros132: Here as well I would suggest using the `` characters to make it simpler to read with not as much string concatenation.

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+        if (isYesterday)
+            return "yesterday";
+        const months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun",
+                        "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+        const monStr = months[then.getMonth()];
+        const dayStr = String(then.getDate());
+        if (then.getFullYear() === nowDate.getFullYear())
+            return monStr + " " + dayStr;
+        return monStr + " " + dayStr + ", " + then.getFullYear();
+    }
```
spiros132, line 1022: Shouldn't all of this be in the translation files?

clipboard/Main.qml
```diff
@@ -0,0 +1,1200 @@
...
+    // does not create missing parent directories, so a fresh install with
+    // no settings.json yet would otherwise fail silently on first pin.
+    // `mkdir -p` is idempotent and costs nothing when the directory
+    // already exists (the usual case — settings.json is already there).
+    onPluginApiChanged: {
+        if (pluginApi) {
+            root.refresh();
+            const home = Quickshell.env("HOME") || "";
+            if (home) {
+                pinnedDirProc.command = ["mkdir", "-p", home + "/.config/noctalia/plugins/clipboard"];
```
spiros132: Here as well, prefer to use the cache instead of the config directory.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+    // closePanelTimer. Uses the same deferred-close idiom as the
+    // delegate click path so toast notifications survive the close.
+    function activateSelection() {
+        const n = root.filteredItems.length;
+        if (n === 0)
+            return;
+        const idx = root.selectedIndex >= 0 ? root.selectedIndex : 0;
+        const entry = root.filteredItems[idx];
+        if (!entry || !entry.id) {
+            Logger.w("clipboard", "activateSelection: no entry at index", idx);
```
spiros132: I would suggest using the name "Clipboard Plugin" since the name "Clipboard" might sound weird when looking at the logs.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+    // If the user presses Delete rapidly on successive rows, each
+    // press overwrites _pendingClampIndex with the current
+    // selectedIndex — which is what we want: focus tracks the
+    // most-recent prune target, not the first one in a burst.
+    function deleteSelection() {
+        if (root.selectedIndex < 0)
+            return;
+        const entry = root.filteredItems[root.selectedIndex];
+        if (!entry || !entry.id) {
+            Logger.w("clipboard", "deleteSelection: no entry at index", root.selectedIndex);
```
spiros132: Here as well, and on every log if it's possible.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                            baseSize: Style.baseWidgetSize * 0.8
+                            // Disable while empty — hides a no-op affordance and
+                            // matches the "there's nothing to clear" state.
+                            enabled: {
+                                const _rev = root.pluginApi?.mainInstance?.itemsRevision ?? 0;
+                                const items = root.pluginApi?.mainInstance?.items ?? [];
+                                return items.length > 0;
+                            }
+                            onClicked: {
+                                root.pluginApi?.mainInstance?.wipe();
```
spiros132, line 843: Maybe a good idea for the future to have a confirmation screen before wiping the whole history.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                            baseSize: Style.baseWidgetSize * 0.8
+                            onClicked: {
+                                var screen = root.pluginApi?.panelOpenScreen;
+                                // Drop the panel's Exclusive keyboard grab so the
+                                // settings popup window (a separate Wayland surface)
+                                // can receive keyboard events. WlrKeyboardFocus.Exclusive
+                                // on the panel surface swallows all input at the
+                                // compositor level, blocking the popup entirely.
+                                if (PanelService.openedPanel)
+                                    PanelService.openedPanel.exclusiveKeyboard = false;
```
spiros132: Is this needed?

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                                if (popupWin) {
+                                    var restore = function() {
+                                        if (!popupWin.hasDialog) {
+                                            if (PanelService.openedPanel)
+                                                PanelService.openedPanel.exclusiveKeyboard = true;
+                                            popupWin.hasDialogChanged.disconnect(restore);
+                                        }
+                                    };
+                                    popupWin.hasDialogChanged.connect(restore);
+                                }
```
spiros132: What is all of this functionality doing here? I believe just calling the `BarService.openPluginSettings` should work.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                            return (pluginApi?.mainInstance?.items?.length ?? 0) > 0;
+                        }
+                        currentIndex: root.currentTypeFilter
+                        tabHeight: Style.toOdd(Style.baseWidgetSize * 0.8)
+                        spacing: Style.marginXS
+                        distributeEvenly: true
+
+                        NTabButton {
+                            tabIndex: 0
+                            text: root.pluginApi?.tr("panel.filter-text") + " (" + root.countForType(0) + ")"
```
spiros132: Would you be able to use translation interpolation instead of string concatenation here?

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                            checked: typeFilterBar.currentIndex === 0
+                            onClicked: root.currentTypeFilter = 0
+                            pointSize: Style.fontSizeXS
+                        }
+                        NTabButton {
+                            tabIndex: 1
+                            text: root.pluginApi?.tr("panel.filter-images") + " (" + root.countForType(1) + ")"
```
spiros132: Here as well

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                            checked: typeFilterBar.currentIndex === 1
+                            onClicked: root.currentTypeFilter = 1
+                            pointSize: Style.fontSizeXS
+                        }
+                        NTabButton {
+                            tabIndex: 2
+                            text: root.pluginApi?.tr("panel.filter-files") + " (" + root.countForType(2) + ")"
```
spiros132: Here as well if it's possible.

clipboard/Panel.qml
```diff
@@ -0,0 +1,1237 @@
...
+                    policy: ScrollBar.AsNeeded
+                }
+            }
+
+            // Image grid (Images tab) -----------------------------------------
+            // Two-column grid so thumbnails use horizontal space efficiently.
+            // Each cell is sized to a 0.65 aspect ratio (landscape-friendly).
+            // Delegates use compact: true so ClipboardItem renders as an image
+            // fill with a hover-reveal delete overlay instead of a text row.
+            GridView {
```
spiros132: Would you be able to use noctalias `NGridView` instead of the qml `GridView`? Less boilerplate and also more consistent with the rest of the shell.

spiros132 (commented):

clipboard/README.md
yanekyuk: Oops, that's for developers to try things out locally. Better to tag it like that.

clipboard/i18n/en.json
yanekyuk: Will sync the missing keys to all locale files.

clipboard/README.md
yanekyuk: What should it look like? Happy to fix once I know the expected format.

clipboard/ClipboardItem.qml
yanekyuk: Will switch to `MouseArea`.

clipboard/Main.qml
yanekyuk: Could you point me to a reference plugin that uses the cache path? Happy to align if that's the established convention.

clipboard/Main.qml
yanekyuk: Same question as above.

clipboard/Main.qml
yanekyuk: Will add the error log.

clipboard/Main.qml
yanekyuk: Same on cache path. Will also extract to a `readonly property`.

clipboard/Main.qml
yanekyuk: Good point, will use template literals throughout.

clipboard/Main.qml
yanekyuk: Will update.

clipboard/Main.qml
yanekyuk: Will update.

clipboard/Main.qml
yanekyuk: Will update.

clipboard/Main.qml
yanekyuk: Will update.

clipboard/Main.qml
yanekyuk: You're right, will move to i18n.

clipboard/Main.qml
yanekyuk: Same as the cache path question above.

clipboard/Panel.qml
yanekyuk: Will update to "Clipboard Plugin".

clipboard/Panel.qml
yanekyuk: Will update across all logs.

clipboard/Panel.qml
yanekyuk: Noted, will add a confirmation dialog in a follow-up.

clipboard/Panel.qml
yanekyuk: Yes — this releases the panel's exclusive keyboard grab before the settings dialog opens. Without it, the panel surface holds exclusive keyboard focus at the Wayland compositor level, which blocks the settings popup (a separate Wayland surface) from receiving any keyboard input. The comment on line 857 explains it in more detail.

clipboard/Panel.qml
yanekyuk: The `BarService.openPluginSettings` call is already there (line 867), but the surrounding code handles a Wayland keyboard focus issue — the panel holds an exclusive keyboard grab that blocks the settings popup from receiving input. The block releases the grab, opens settings, then restores the grab when the popup closes. If the shell handles this internally I'm happy to simplify, but without it the settings dialog isn't keyboard-accessible when opened from the panel.

clipboard/Panel.qml
yanekyuk: Will use translation interpolation instead.

clipboard/Panel.qml
yanekyuk: Will update.

clipboard/Panel.qml
yanekyuk: Will update.

clipboard/Panel.qml
yanekyuk: Will replace with `NGridView`.

clipboard/README.md
spiros132: It's completely up to you to find the format that you prefer. It might be better for users installing this plugin for it to say something like "go to settings->plugins and install it from there" since that's how most people will install it. If you'd like you can also maybe add a section that's called "Development" where in there it says how to develop the plugin

clipboard/Main.qml
spiros132: If I'm not wrong the assistant panel plugin uses the cache directory. Also the video wallpaper plugin but not specifically the plugins directory of it.

clipboard/Panel.qml
spiros132: Try it out and see if it works correctly, there is supposed to be only one popup open at the same time so having two different panels open shouldn't be happening. Try to not include this and use the plugin and see if it works as it should.

clipboard/Main.qml
yanekyuk: Fixed — added a `dataDir` readonly property pointing to `~/.cache/noctalia/plugins/clipboard` and derived all file paths from it.

clipboard/Main.qml
yanekyuk: Fixed, same as above.

clipboard/Main.qml
yanekyuk: Fixed — added a `dataDir` readonly property pointing to `~/.cache/noctalia/plugins/clipboard` and derived all file paths from it.

clipboard/Main.qml
yanekyuk: Fixed, same as above.

clipboard/Main.qml
yanekyuk: Fixed — also extracted to the shared `dataDir` property.

clipboard/Main.qml
yanekyuk: Fixed, same as above.

clipboard/README.md
yanekyuk: Updated — Installation now points to Settings → Plugins. The git clone instructions are moved to a Development section.

clipboard/Main.qml
yanekyuk: Done — applied template literals for commands where the interpolated value is a validated numeric id. Commands that handle user-controlled content (file URIs, clipboard text, JSON payload) still use the $1/$2 positional pattern to keep shell metacharacters inert.

clipboard/Main.qml
yanekyuk: Done.

clipboard/Main.qml
yanekyuk: Done.

clipboard/Main.qml
yanekyuk: Done.

clipboard/Main.qml
yanekyuk: Done.

clipboard/ClipboardItem.qml
yanekyuk: Done — removed the separate `HoverHandler` and added `hoverEnabled: true` to the existing `rowArea` MouseArea, referencing `rowArea.containsMouse` throughout.

clipboard/Panel.qml
yanekyuk: Why is "there is supposed to be only one popup open at the same time"? I kind of find it annoying when you change settings and don't see the result directly.

spiros132: I'm sorry but I'm a bit confused on what you mean. In the shell itself there can only be one panel open at the same time. So if you have a plugin panel open, and try to open the settings, the plugin panel will close automatically since only one can be open at the same time. Here is the relevant [source code](https://github.com/noctalia-dev/noctalia-shell/blob/main/Services/UI/PanelService.qml#L238) if you want to see.

clipboard/manifest.json
```diff
@@ -0,0 +1,33 @@
...
+  "author": "yanekyuk",
+  "repository": "https://github.com/noctalia-dev/noctalia-plugins",
+  "license": "MIT",
+  "description": "Clipboard history panel with search, pinning, and image previews.",
+  "tags": [
+    "Utility",
+    "Bar",
+    "Panel",
+    "Niri",
+    "Hyprland"
```
spiros132: Is the plugin specific for Niri or Hyprland only?

clipboard/ClipboardItem.qml
```diff
@@ -0,0 +1,665 @@
...
+    // Tracks whether the row is currently being pressed (mouse down but not
+    // yet released). Driven by rowArea.onPressed/onReleased below.
+    property bool pressed: false
+
+    // HoverHandler is geometry-based and not affected by z-ordering of child
+    // MouseAreas (e.g. NIconButton internals). Using it instead of
+    // rowArea.containsMouse prevents the flicker loop where hovering over a
+    // button causes rowArea (z:-1) to lose hover → buttons hide → rowArea
+    // regains hover → buttons show → repeat.
+    HoverHandler { id: cardHover }
```
spiros132, line 227: Since you already have a `MouseArea`, is there a reason to have this `HoverHandler`?

clipboard/Main.qml
spiros132: This isn't resolved yet.

clipboard/Main.qml
```diff
@@ -0,0 +1,1196 @@
...
+            // managers we care about here; paste into a text target
+            // still works because most text widgets also accept
+            // x-special as plain text.
+            //
+            // The URI is passed via $1 (not interpolated) so shell
+            // metacharacters in the path cannot execute. printf's
+            // format string contains only constants so "copy\n" and
+            // the trailing newline are emitted verbatim.
+            copyProc.command = ["bash", "-c",
+                'printf "copy\n%s\n" "$1" | wl-copy --type x-special/gnome-copied-files', "--", uri];
```
spiros132, line 688: Same thing here as well, prefer to put the variable inside the string itself.

clipboard/Main.qml
spiros132: Still isn't resolved

spiros132 (commented):
Some more feedback about the PR! :D

clipboard/ClipboardItem.qml
yanekyuk: No change. `card` is the semantic name used throughout the file and children reference `card.*`. Renaming to `root` would lose that meaning with no practical benefit.

clipboard/Main.qml
yanekyuk: Reversing the previous response. `preview` is raw clipboard content: it can contain quotes, dollar signs, backticks, newlines, any shell metacharacter. Interpolating it into a template literal would inject it directly into the shell string, which is exploitable. The `$1` argv pattern passes `preview` as a positional argument that bash never interprets as shell syntax. This is intentional and the inline comment explains it. No change.

clipboard/Main.qml
yanekyuk: Walking back the previous response. These are formatting tokens in a JS utility function, not UI strings. Putting each month abbreviation as a separate i18n key is unusual and harder to maintain. If locale-aware dates are a requirement, the right tool is `Intl.DateTimeFormat`, not JSON keys. No change.

clipboard/Panel.qml
yanekyuk: The keyboard grab code has been removed. The current implementation is just `BarService.openPluginSettings(root.pluginApi?.panelOpenScreen, root.pluginApi?.manifest)`, nothing else.

clipboard/manifest.json
yanekyuk: Removed the Hyprland tag. The plugin was developed and tested on Niri only.

clipboard/Main.qml
yanekyuk: This is already a template literal: `cliphist decode ${sid} | wl-copy`. `sid` is validated as `[0-9]+` before this point.

clipboard/manifest.json
```diff
@@ -0,0 +1,32 @@
...
+  "minNoctaliaVersion": "4.1.2",
+  "author": "yanekyuk",
+  "repository": "https://github.com/noctalia-dev/noctalia-plugins",
+  "license": "MIT",
+  "description": "Clipboard history panel with search, pinning, and image previews.",
+  "tags": [
+    "Utility",
+    "Bar",
+    "Panel",
+    "Niri"
```
spiros132: Usually having the Niri tag means that it only runs with Niri, I believe you don't need it since it's a clipboard which doesn't have anything compositor specific.

spiros132 (commented):
Just one last thing, otherwise I believe it's ready to get merged in. :)

Cleboost (approved):
lgtm

clipboard/manifest.json
yanekyuk: I only tested this in Niri, so I can only confirm that it works with Niri. Leaving it in for now.

spiros132: Once again, the tag doesn't have to do with what it's tested against but what it works in. Since this isn't compositor specific it doesn't need to have the Niri tag.

yanekyuk: Alright. Since `cliphist` and `wl-copy` are generic Wayland utilities, It makes sense to remove Niri from the list.

spiros132 (commented):

PR #662, opened 2026-04-16
feat(github-feed): add support for self-hosted GitHub URLs and update settings UI

pehlicd:
### Changes
- Added support for self-hosted GitHub urls.
- Added translations with the help of AI.

### Motivation
I would like to be able configure github url to fetch feed from my self hosted GitHub.


spiros132 (commented):
Since @linuxmobile hasn't responded for quite a while and I believe everything looks good, I'll merge this in :)

PR #663, opened 2026-04-16
feat(mimeapp-gui): add ControlCenterWidget

neyfua:


Mathew-D:
@neyfua What are you trying to add here? The BarWidget.qml basically already has this.

neyfua:
> @neyfua What are you trying to add here? The BarWidget.qml basically already has this.

I'm trying to add it as a Control Center Widget for people that doesn't want plugin widgets on their bar but still need to use it, it could be added inside Control Center instead

image

Mathew-D:
Makes sense to me and looks good.

neyfua:
> Makes sense to me and looks good.

Thanks :P

@Ly-sec if u have some times

Mathew-D:
You might want to add to the readme about the Control Center part you added.

neyfua:
> You might want to add to the readme about the Control Center part you added.

Alright, I forgor

PR #664, opened 2026-04-16
feat(screen-shot-and-record): Added Sway

Mathew-D:
Adds sway support with window detection
Adds mangoWC support with just area selection
Add setting of where to save video and images
Still No Niri support. Added a block to tell them it doesn't work.
Common Objects that gets extended for window detection in Sway and Hyrland


Pulsar33550336 (approved):
Looks good, I don't have time to test it right now, but from the source code perspective, there shouldn't be any issues.

Pulsar33550336:
A suggestion: move `ScreenShot{Sway,Hypr}.qml` to a separate folder (like `Platform`)?

Mathew-D:
@Pulsar33550336 if I try and move them I get a lot of error about it not being able to find the Type that I make in ScreenShot.  If it is OK with you I would really like to leave them in the root as I dont know how to fix that problem.

Pulsar33550336:
It's OK @Mathew-D

Mathew-D:
@Ly-sec if you have some time

PR #665, opened 2026-04-17
feat(monitor-layout): Allows gui setup for multimonitors for sway and hyprland

Mathew-D:
Allows you to setup monitor placements in Sway and Hyprland.
Gives you the settings you can copy and paste into your config files if you want them to stay after log out.
Gave directions if someone wants to add other backends.

Should solve https://github.com/noctalia-dev/noctalia-shell/issues/2387****


monitor-layout/Main.qml
```diff
@@ -0,0 +1,353 @@
...
+        root.errorText = root.applyStderr !== "" ? root.applyStderr : root.t("errors.applyFailed");
+        root.statusText = "";
+        Logger.e("MonitorLayout", root.errorText);
+      }
+    }
+  }
+
+  function t(key, interpolations) {
+    return pluginApi ? pluginApi.tr(key, interpolations) : key;
+  }
```
spiros132: This isn't needed, just use the regular `pluginApi?.tr` function and it will automatically return the key if the key is missing.

monitor-layout/Main.qml
```diff
@@ -0,0 +1,353 @@
...
+  //
+  //   buildConfigFileContent(outputs) -> { content } | { error }
+  //     Build a config file snippet for the given outputs (for copy-to-config).
+  //     Return { error: string } if the config cannot be built.
+  //
+  // To add a new backend:
+  //   1. Create backends/MyBackend.js implementing the interface above.
+  //   2. Add: import "backends/MyBackend.js" as MyBackend  (top of file)
+  //   3. Add a case for your backend id in each function below.
+  // ---
```
spiros132, line 156: Really good documentation, might be a good idea to also document what the `outputs` mean, what kind of fields should the object have? Or what kind of object is `content` or `script` for example.

monitor-layout/Main.qml
```diff
@@ -0,0 +1,353 @@
...
+
+  function backendBuildConfigFileContent() {
+    if (backendId === "sway") return SwayBackend.buildConfigFileContent(draftOutputs);
+    if (backendId === "hyprland") return HyprlandBackend.buildConfigFileContent(draftOutputs);
+    return { "error": t("errors.unsupportedBackend") };
+  }
+
+  function snapToGridEnabled() {
+    return cfg.snapToGrid ?? defaults.snapToGrid ?? true;
+  }
```
spiros132: Prefer to use a binding instead of a function. For example:
```qml
readonly property bool example: cfg.example ?? defaults.example ?? true
```

monitor-layout/Panel.qml
```diff
@@ -0,0 +1,830 @@
...
+  readonly property real sceneScale: computeSceneScale()
+  property string editPosX: ""
+  property string editPosY: ""
+  property string editScale: ""
+
+  anchors.fill: parent
+
+  function t(key, interpolations) {
+    return pluginApi ? pluginApi.tr(key, interpolations) : key;
+  }
```
spiros132: Here as well, this function isn't needed.

monitor-layout/Panel.qml
```diff
@@ -0,0 +1,830 @@
...
+  function layoutOutputHeight(output) {
+    return output && output.logicalHeight ? output.logicalHeight : output.height;
+  }
+
+  function resolutionSummary(output) {
+    if (!output) {
+      return "";
+    }
+
+    var scaleText = output.scale && output.scale !== 1 ? "  |  " + t("panel.scale") + ": " + output.scale : "";
```
spiros132: Would you be able to use translation interpolation here?

monitor-layout/Panel.qml
```diff
@@ -0,0 +1,830 @@
...
+    return model;
+  }
+
+  function escapeForShell(text) {
+    return String(text || "").replace(/'/g, "'\\''");
+  }
+
+  function copyToClipboard(text) {
+    var escaped = escapeForShell(text);
+    var cmd = "printf '%s' '" + escaped + "' | wl-copy || printf '%s' '" + escaped + "' | xclip -selection clipboard || printf '%s' '" + escaped + "' | xsel --clipboard --input";
```
spiros132: Would you be able to use the `` characters to make this a bit cleaner? For example:
```qml
var cmd = `print ... "${escaped}" || ... "${escaped}"`
```

monitor-layout/Panel.qml
```diff
@@ -0,0 +1,830 @@
...
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+
+            ColumnLayout {
+              anchors.fill: parent
+              anchors.margins: Style.marginM
+              spacing: Style.marginM
+
+              NText {
+                text: t("panel.canvasTitle")
```
spiros132: As before, on all of these don't use your own `t` function, prefer to use the `pluginApi?.tr` function

monitor-layout/Panel.qml
```diff
@@ -0,0 +1,830 @@
...
+                  }
+
+                  NText {
+                    text: t("panel.transform") + ": " + (selectedOutput ? selectedOutput.transform : "")
+                    pointSize: Style.fontSizeS
+                    color: Color.mOnSurfaceVariant
+                  }
+
+                  NText {
+                    text: t("panel.backend") + ": " + (mainInstance ? mainInstance.backendId : "")
```
spiros132: Would you be able to use translation interpolations on these two as well?

spiros132 (commented):

monitor-layout/Main.qml
Mathew-D: Added what outputs is above in the comment

spiros132 (commented):
It looks great, thank you so much for the PR! :)

PR #667, opened 2026-04-17
Add management URL configuration and update manifest properties

Ebnater:
Adds support for a configurable Admin Console URL in the NetBird plugin, with automatic URL detection from NetBird status output.

Also resolves the final URL with a clear priority: user-defined setting first, auto-detected management URL second, and default fallback last.

Changes:

- Add support for storing a custom managementUrl setting in `manifest.json` and `Settings.qml`.
- Update status parsing in `Main.qml` to detect management URL fields from NetBird JSON output.
- Add URL normalization in `Main.qml` so host-only values are converted into valid URLs.
- Update the Admin Console button in `Panel.qml` to open the resolved URL instead of a hardcoded one.
- Add new translation keys for the settings field in `en.json`, `de.json`, and `fr.json`

### DIFF
--- a/netbird/Main.qml
+++ b/netbird/Main.qml
@@ -29,6 +29,15 @@ Item {
     property bool compactMode: _computeCompactMode()
     property bool showIpAddress: _computeShowIpAddress()
     property bool showPing: _computeShowPing()
+    property string configuredManagementUrl: _computeConfiguredManagementUrl()
+    readonly property string adminConsoleUrl: {
+        var configured = normalizeManagementUrl(configuredManagementUrl);
+        if (configured !== "")
+            return configured;
+        if (managementUrlDetected !== "")
+            return managementUrlDetected;
+        return normalizeManagementUrl(pluginApi?.manifest?.metadata?.defaultSettings?.managementUrl || "https://app.netbird.io/");
+    }
 
     function _computeRefreshInterval() {
         return pluginApi?.pluginSettings?.refreshInterval ?? 5000;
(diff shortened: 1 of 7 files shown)


spiros132:
Would you be able to fix the automatic code quality review as well?

spiros132:
@Cleboost could you review this PR? :)

netbird/manifest.json
Cleboost:
```suggestion
  "version": "1.1.0",
```

netbird/Main.qml
```diff
@@ -63,6 +115,8 @@ Item {
     property var peerList: []
     property bool managementConnected: false
     property bool signalConnected: false
+    property string managementFqdn: ""
```
Cleboost: why do you need this variable? tt's never read and seems to be a duplicate of `managementUrlDetected`

Ebnater: My bad :)

netbird/Main.qml
```diff
@@ -42,12 +51,55 @@ Item {
...
+
+        if (url.indexOf("://") === -1) {
+            if (url.indexOf("localhost") === 0 || url.indexOf("127.0.0.1") === 0 || url.indexOf("[::1]") === 0) {
+                return "http://" + url;
+            }
+            return "https://" + url;
+        }
+
+        return url;
+    }
```
Cleboost: i'm not sure but i think it can be better

```suggestion
	function normalizeManagementUrl(rawUrl) {
	  const url = String(rawUrl || "").trim();
	  if (!url) return "";
	
	  if (/^https?:\/\//i.test(url)) return url;
	
	  const isLocal = /^(localhost|127\.0\.0\.1|\[::1\])(?::|\/|$)/i.test(url);
	  return `${isLocal ? "http" : "https"}://${url}`;
	}
```

Cleboost (approved):
lgtm <3 thank for your pr

PR #672, opened 2026-04-18
obs-control: more reliable OBS process search

bpavuk:
NixOS wraps OBS into `.obs-wrapper`, which `pgrep -x obs` fails to find. since both `.obs-wrapper` and the `obs` command on other Linux distros run the same `obs` binary, `pidof` works more reliably.

I spent more time laying out the reasoning here than fixing this...


bpavuk:
GitHub's built-in editor is f'ed up. please, review #673 instead

PR #673, opened 2026-04-18
fix(obs-control): use `pidof` instead of `pgrep`

bpavuk:
NixOS wraps OBS into `.obs-wrapper`, which `pgrep -x obs` fails to find. since both `.obs-wrapper` and the `obs` command on other Linux distros run the same `obs` binary, `pidof` works more reliably

I spent more time laying out the reasoning here than fixing this...

---

UPD: for some reason GitHub decided to rename this branch to `patch-1` although I gave it a different name. closing #672 and reopening here.

### DIFF
--- a/obs-control/Main.qml
+++ b/obs-control/Main.qml
@@ -858,7 +858,7 @@ Item {
   Process {
     id: obsProbeProcess
     running: false
-    command: ["pgrep", "-x", "obs"]
+    command: ["pidof", "obs"]
 
     onExited: function(exitCode) {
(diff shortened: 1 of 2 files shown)


Cleboost (changes requested):
Don't forget to bump patch version in `manifest.json`

spiros132:
Since @ayagmar hasn't responded in a few days and this is a very simple fix I'll merge this in :)

ayagmar:
ah sorry, thanks for the fix!

PR #680, opened 2026-04-18
MimeApp GUI

neyfua:
@Mathew-D sorry for the sudden of alot of changes on the MimeApp GUI plugin like this but below is the brief change logs for the plugin and hope that it'll help with the user's experience more

### Change logs for MimeApp GUI:
1. Added `Settings.qml` with options to change BarWidget icon colors and copy IPC command directly inside the settings panel
2. Improve the BarWidget icon by removing the uneccessary background behind the icon and to follow the development guidelines of Noctalia plugin creation
3. Change BarWidget icon from "file-text" to "file-dots" to avoid having the same icon type with `notes-scratchpad` plugin
4. Changed the IPC command from "open" to "openPanel" for a more clarification on the IPC command's purpose
5. Added some new translations for the new stuff added above
6. Changed the default icon color from mPrimary to mOnSurface and users have option to choose their own icon color inside settings panel
7. Added default metadata settings of the icon color inside `manifest.json` and bumping version up

### DIFF
--- a/mimeapp-gui/BarWidget.qml
+++ b/mimeapp-gui/BarWidget.qml
@@ -1,26 +1,63 @@
 import Quickshell
 import qs.Commons
+import qs.Services.UI
 import qs.Widgets
 
 NIconButton {
-  id: root
+	id: root
 
-  property var pluginApi: null
-  property ShellScreen screen
-  property string widgetId: ""
-  property string section: ""
-  property int sectionWidgetIndex: -1
-  property int sectionWidgetsCount: 0
+	property var pluginApi: null
 
-  readonly property string screenName: screen?.name ?? ""
+	property ShellScreen screen
+	property string widgetId: ""
+	property string section: ""
+	property int sectionWidgetIndex: -1
+	property int sectionWidgetsCount: 0
 
-  baseSize: Style.getCapsuleHeightForScreen(screenName)
-  applyUiScale: false
-  customRadius: Style.radiusL
-  icon: "file-text"
-  tooltipText: pluginApi?.tr("bar.tooltip")
+	property var cfg: pluginApi?.pluginSettings || ({})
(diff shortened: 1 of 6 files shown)


neyfua:
> ### Automatic Code Quality Review
> ### File: mimeapp-gui/Settings.qml
> 
>     * [(H)](680) Line 49: Use translations instead of hardcoded text. Instead of: `"Example Label"`. To: `pluginApi?.tr("panel.example-label")`
> 
> 
> ```qml
> +				text: "qs -c noctalia-shell ipc call plugin:mimeapp-gui openPanel"
> ```

Not needed since that's just a command and it works all the same on the other languages

Mathew-D:
I don't have access to my testing machine for a day or two but looking over the code I did have questions about the use of Rectangle and TextEdit,  If you look at the AGENTS.md it asks use to stay away from raw Qt types and to use Noctalia widgets instead.  Could they be switched?

neyfua:
> I don't have access to my testing machine for a day or two but looking over the code I did have questions about the use of Rectangle and TextEdit, If you look at the AGENTS.md it asks use to stay away from raw Qt types and to use Noctalia widgets instead. Could they be switched?

I took those from other plugins so that's my bad :), I'll remove it

neyfua:
@Mathew-D I just checked again with Lysec, it's okay to use Rectangle and TextEdit since it's for the purpose of adding an IPC command inside the Settings UI panel, and I can't find a equivalent way to do it like this with the available widgets of Noctalia. So if it's okay for u then we can keep it, if not I'll remove the entire IPC command inside Settings UI panel part.

Mathew-D:
@neyfua if Lysec is good with their use so am I.  I looked though all code changes and everything looks good.  @Ly-sec  if you have some time I am good with this.

PR #694, opened 2026-04-19
[Wallcards] Added settings panel with some toggleable settings

elrondforwin:
Nothing changed with the default look, just added toggles.

image

### DIFF
--- a/wallcards/Settings.qml
+++ b/wallcards/Settings.qml
@@ -1,0 +1,51 @@
+import QtQuick
+import QtQuick.Layouts
+import qs.Commons
+import qs.Widgets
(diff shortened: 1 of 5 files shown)


tonigineer:
Hola, 

first of all, thanks for adding to the plugin. I have already implemented a settings panel and other stuff, but didn't finalize and publish it yet. Private life is coming in my way :)

If this is needed right now, go ahead and merge it. But I would force push my changes later. Hopefully, I can add it in about a weeks time or two. 

FYI, I did not have a look at the code. Just wanted to give a quick reply.

Best regards,
Toni

PR #697, opened 2026-04-19
chore: ignore CLAUDE.local.md

kohanyirobert:
## Summary

- Adds `CLAUDE.local.md` to `.gitignore` so local Claude Code project instructions are not accidentally committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@
 # OS
 .DS_Store
 Thumbs.db
+
+# AI / agents
+CLAUDE.local.md


adriamartin91:
This is specific to your local setup, so I don’t think it belongs in the project..

kohanyirobert:
> This is specific to your local setup, so I don’t think it belongs in the project..

I can interpret this in two ways. My feeling is that you're suggesting this change *not* to be merged, am I right?
Why? That's the whole point of `.gitignore`, isn't it? To ignore files not related to a project, that would generate noise in `git status`. `CLAUDE.local.md` does not belong to any project so let's ignore it.

Where's the harm?

How are the following ignored patterns different? VS Code is specific to a local setup, also JetBrains products, also Vim and Emacs.

```
# IDE / editor
.vscode/
.idea/
*.swp
*.swo
*~
```

kohanyirobert:
👍

PR #702, opened 2026-04-20
Improve USB drive manager eject reliability and bar icon settings

LyHug:
## Summary
- make auto-mount only target newly detected, currently unmounted devices
- queue mount/eject actions to avoid overlapping `udisksctl` operations
- safely eject the whole removable device by unmounting every mounted sibling volume before power-off
- add a configurable bar icon setting with icon picker support
- persist icon color changes immediately and document the new behavior

## Why
The current plugin can show an `Eject failed` notification even after a normal-looking eject flow because it powers off the parent disk even when another partition on the same device is still mounted. In practice this also interacts badly with auto-mount, which can re-trigger mounts right after unmount/eject-driven refreshes.

## Notes
- JSON files were validated with `python -m json.tool`.
- `qmllint` is not available in this environment, so I could not run it here.


Cleboost (approved):
lgtm <3

usb-drive-manager/BarWidget.qml
```diff
@@ -38,21 +43,23 @@ NIconButton {
     property var cfg: pluginApi?.pluginSettings || ({})
     property var defaults: pluginApi?.manifest?.metadata?.defaultSettings || ({})
     readonly property string iconColorKey: cfg.iconColor ?? defaults.iconColor ?? "none"
-    readonly property color iconColor: Color.resolveColorKey(iconColorKey)
+    readonly property bool hasCustomIconColor: iconColorKey !== "" && iconColorKey !== "none"
+    readonly property color mountedBgColor: Color.mPrimaryContainer || Color.mPrimary || Style.capsuleColor
+    readonly property color mountedFgColor: Color.mOnPrimaryContainer || Color.mOnPrimary || Color.mOnSurface
```
spiros132, line 48: The `Color` singleton will never be null or undefined. So I don't see why there's a need for fallback values.

usb-drive-manager/DeviceCard.qml
```diff
@@ -127,14 +135,14 @@ Rectangle {
                 Layout.fillWidth: true
                 height: Style.marginXS
                 radius: height / 2
-                color: Color.mOutlineVariant
+                color: Color.mOutlineVariant || Qt.alpha(Color.mOutline, 0.35)
```
spiros132, line 138: Same thing here, I don't see what the fallback value does.

usb-drive-manager/DeviceCard.qml
```diff
@@ -127,14 +135,14 @@ Rectangle {
...
 
                 Rectangle {
                     width: parent.width * Math.min((device?.usedPercent ?? 0) / 100, 1)
                     height: parent.height
                     radius: parent.radius
                     color: (device?.usedPercent ?? 0) > 90 ? Color.mError
-                         : (device?.usedPercent ?? 0) > 75 ? Color.mWarning
+                         : (device?.usedPercent ?? 0) > 75 ? (Color.mWarning || Color.mTertiary || Color.mSecondary || Color.mPrimary)
```
spiros132, line 145: Same thing here.

usb-drive-manager/DeviceCard.qml
```diff
@@ -223,8 +231,8 @@ Rectangle {
                 baseSize: Style.baseWidgetSize * 0.8
                 colorBg: Color.mSurfaceVariant
                 colorFg: Color.mOnSurfaceVariant
-                colorBgHover: Color.mErrorContainer
-                colorFgHover: Color.mOnErrorContainer
+                colorBgHover: Color.mErrorContainer || Qt.alpha(Color.mError, 0.18)
+                colorFgHover: Color.mOnErrorContainer || Color.mOnError || Color.mError
```
spiros132, line 235: And here as well.

usb-drive-manager/manifest.json
spiros132, line 1: Don't forget to also bump the version number.

usb-drive-manager/Main.qml
```diff
@@ -108,11 +110,25 @@ Item {
...
                     root.devicesChanged()
+
+                    if (shouldAutoMount) {
+                        const mountCandidates = internal.getAutoMountCandidates(previousPaths, nextDevices)
+                        if (mountCandidates.length > 0) {
+                            Qt.callLater(() => root.queueMountDevices(mountCandidates))
+                        }
+                    }
                 } catch (e) {
                     console.warn("[usb-drive-manager] Failed to parse lsblk output:", e)
```
spiros132, line 133: Change this so that it uses `Logger` instead of using the `console`.

spiros132: Also if you find any other `console` calls, please change them to use the `Logger`

spiros132 (commented):

PR #712, opened 2026-04-21
latency-monitor: fix NButton background property and add animations

guemidiborhane:


latency-monitor/manifest.json
spiros132, line 1: Don't forget to bump the version number as well.

PR #728, opened 2026-04-21
fix(clipper): ReferenceError teardown + atomic write data-loss hotfix (v2.4.3)

blackbartblues:
## Summary

Two related hotfixes bundled on the same branch, bumped from 2.4.1 to **2.4.3**.

### 1. ReferenceError on teardown (was v2.4.2)

Fires in `Main.qml` `Component.onDestruction` on every panel/shell teardown:

```
WARN scene: file:///.../clipper/Main.qml?v=0[1300:-1]: ReferenceError: wlCopyProc is not defined
```

The `wlCopyProc` id does not exist in the file — leftover from an earlier refactor. wl-copy usage already lives in `copyToClipboardProc` and direct `Quickshell.execDetached(["wl-copy", ...])` calls, so no replacement logic is needed — just delete the two dead lines.

### 2. Data-loss hotfix: atomic writes for pinned items and notecards (v2.4.3)

Investigating a silently-vanished notecard revealed a 0-byte ghost file in `notecards/`. Root-caused to three compounding issues in the save path:

1. **Non-atomic shell redirection.** All writes used `base64 -d > "$file"`, which opens the target with `O_TRUNC` **before** decoding and writing. If the process was interrupted between truncate and write (shell restart, kill, OOM, base64 failure), the file was left at 0 bytes with no recovery possible.
2. **Rename race in `updateNoteCard`.** When a note's title changed, the old file was deleted via a parallel `rm` `execDetached` call issued alongside the new `saveNoteCard`. If the `rm` completed but the save failed, the note vanished entirely.
3. **Empty-base64 edge case.** If `stringToBase64()` returned an empty string, the shell pipeline still truncated the target, producing a guaranteed 0-byte file.

**Fix:** new `atomicWriteBase64(filePath, base64, oldFilePath)` helper in `Main.qml`:

- Writes to a temp file (`<file>.tmp`) first, verifies non-empty (`[ -s "$t" ]`), then atomically renames over the target with `mv -f`.
- Only removes the old file **after** the new file is safely in place (and only when the old path differs from the new path).
- Passes `filePath`, `oldFilePath`, and `base64` via argv (not string interpolation) — shell-injection safe and robust against filenames with spaces/quotes.
- Refuses empty writes up front (`base64.length === 0`) and logs a warning.
- `saveNoteCard` validates `note.id` and `JSON.stringify` length >= 10 before writing.

Applied to `savePinnedFile`, `saveNoteCard`, `updateNoteCard`, and `exportNoteCard`. `Component.onCompleted` now sweeps stale `*.json.tmp` files from prior interrupted writes.

## Changes

- `clipper/Main.qml`: remove dead `wlCopyProc` refs + add `atomicWriteBase64` helper + refactor all four save sites + stale-tmp sweep
- `clipper/manifest.json`: bump version 2.4.1 → 2.4.3
- `clipper/CHANGELOG.md`: add v2.4.2 and v2.4.3 entries

## Test plan

- [x] `Main.qml` no longer references any undefined ids (grep)
- [x] wl-copy functionality still routes through `copyToClipboardProc` and `Quickshell.execDetached`
- [x] Open/close panel with `qs -c noctalia-shell` and confirm the ReferenceError no longer appears in logs
- [x] Create a notecard with Polish diacritics / emoji in the title, edit the title, then close the shell — verify no 0-byte `.json` files appear in `~/.config/noctalia/plugins/clipper/notecards/`
- [x] Pin several clipboard entries, then kill the shell mid-save — verify `pinned.json` is either the old valid content or the new valid content, never 0 bytes

### DIFF
--- a/clipper/CHANGELOG.md
+++ b/clipper/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Clipper Plugin - Comprehensive Changelog
 
+## Version 2.4.2 (2026-04-22)
+
+### Fix ReferenceError on plugin teardown
+
+- Removed a dead `wlCopyProc` reference in `Component.onDestruction` (`Main.qml`) that threw `ReferenceError: wlCopyProc is not defined` when the panel or shell was torn down. The id was left over from an earlier refactor; wl-copy usage already lives in `copyToClipboardProc` and direct `Quickshell.execDetached(["wl-copy", ...])` calls, so no replacement logic is needed.
+
 ## Version 2.4.1 (2026-04-21)
 
 ### Fix Qt.btoa deprecation warnings
--- a/clipper/CHANGELOG.md
+++ b/clipper/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Clipper Plugin - Comprehensive Changelog
 
+## Version 2.4.3 (2026-04-22)
+
+### Data-loss hotfix: atomic writes for pinned items and notecards
+
+Fixes a data-loss bug that could silently destroy pinned items (`pinned.json`) or individual notecards (`notecards/*.json`) under rare timing conditions. A 0-byte ghost file left on disk was traced back to three compounding issues in the save path:
+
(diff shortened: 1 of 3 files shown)


blackbartblues:
Don't merge! I will commit Very fast fix because 2.4.1 merged with one big bug and it can clean someones notecards!.

blackbartblues:
Merge please!

PR #729, opened 2026-04-22
feat(translator): add configurable DeepLX-like backend

RanderDouble:
## Summary

  This PR adds a minimal `DeepLX-like API` backend option to the Translator plugin.

  ## Changes

  - add a new `DeepLX-like API` backend in plugin settings
  - add configurable `API URL` and optional `API Key` fields
  - send custom backend translation requests as `POST` JSON
  - attach `Authorization: Bearer <token>` when an API key is provided
  - support parsing either `data` or `translations[0].text` from the response
  - include backend and API URL in the cache key to avoid stale cached results
  after switching backends
  - document DeepLX-like configuration in the plugin README

  ## Motivation

  The current Google backend relies on an unofficial endpoint and may fail due to
  rate limiting or anti-abuse checks.

  This change adds support for self-hosted DeepLX-like endpoints while keeping the
  implementation small and preserving the existing Google and DeepL backends.

  ## Notes

  - existing `google` and `deepl` backends are unchanged
  - for DeepLX instances using query-token auth, the token can be included directly
  in the configured API URL

### DIFF
--- a/translator/LauncherProvider.qml
+++ b/translator/LauncherProvider.qml
@@ -137,12 +137,7 @@ Item {
     }
 
     function getErrorMessage(status, fallback) {
-        if (status === 401) return "Invalid API Key";
-        if (status === 403) return "Forbidden";
-        if (status === 404) return "API endpoint not found";
-        if (status === 429) return "Rate limited";
-        if (status >= 500) return "Upstream service error";
-        return fallback || (pluginApi?.tr("messages.connectionError") || "Connection error");
+        return fallback || pluginApi?.tr("messages.connectionError");
     }
 
     function translateText(text, targetLanguage, cacheKey) {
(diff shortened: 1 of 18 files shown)


RanderDouble:
Addressed the i18n review feedback.

  Changes in this update:
  - removed the new hardcoded `API URL` label
  - removed the new translation fallbacks introduced in the DeepLX-like backend
  changes
  - kept the error handling scoped to the existing generic messages plus
  `missingApiUrl`
  - added localized strings for the new DeepLX-like settings fields and missing URL
  message

Cleboost (approved):
thanks <3

translator/LauncherProvider.qml
```diff
@@ -134,6 +136,10 @@ Item {
         return pluginApi?.pluginSettings?.backend || pluginApi?.manifest?.metadata?.defaultSettings?.backend || "google";
     }
 
+    function getErrorMessage(status, fallback) {
+        return fallback || pluginApi?.tr("messages.connectionError");
+    }
```
spiros132, line 141: I'm a bit confused why this is needed.

spiros132 (commented):
One minor feedback otherwise it lgtm :)

PR #735, opened 2026-04-22
feat(arch-updater): v2.0.0 - new settings UI and launcher updates search

Astinator:
Redesigns the settings UI to be easier to use
Splits up the update logic into system + aur + flatpak, instead of system + flatpak, and sets a source for each update
Adds a launcher command that lists available updates and opens them in their respective repos when selected


spiros132 (commented):
Looks great! :D

PR #738, opened 2026-04-22
fix(screen-shot-and-record): Remove the getaudiooutput

Mathew-D:
remove var in argument --audio="$(getaudiooutput)" with --audio
remove the whole function getaudiooutput() since it is no longer needed, wf-recorder selects the default PipeWire source automatically 
As per https://github.com/noctalia-dev/noctalia-plugins/discussions/737

### DIFF
--- a/screen-shot-and-record/manifest.json
+++ b/screen-shot-and-record/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "screen-shot-and-record",
   "name": "ScreenShot & Record",
-  "version": "1.0.3",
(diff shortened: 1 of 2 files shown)


Mathew-D:
@Atilil as per your post.

Pulsar33550336:
> @Pulsar33550336- 此 PR 将修改您的插件。请在方便时进行审阅。

Done.

Mathew-D:
@Ly-sec when you get a sec please.

PR #745, opened 2026-04-24
Fix/kubectl ctx prfix(kubectl-ctx): Changed preview.png

ultherego:
Changed preview.png with correct sizing


ultherego:
A minor cosmetic change

PR #759, opened 2026-04-25
fix(tailscale): Tailscale icon tweaks

weskoop:
Just a couple small tweaks to the icon handling in the Tailscale plugin


spiros132:
Don't forget to bump the version in the manifest as well :)

PR #763, opened 2026-04-26
Wallcards refactor

tonigineer:
Hola, 

first of all, sorry for this huge PR. I refactored quite a lot compared to the initial version. From now on, fixes and progress should come in way smaller increments.

To the PR itself, I refactored a lot, added a settings page and keybinds as well as video wallpaper via `mpvpaper`. The layout of the cards can be changed live via keybinds and saved to make them persistent. I also added a shortcut overview, which can be toggled (?).

I am still not using the thumbnail functionality from Noctalia (as already proposed to me earlier), due to the thumbnails for videos and the hex code suffix for the dominant color.

That should be it, for now :)

Best regards,
Toni


wallcards/.qmlls.ini
spiros132, line 1: Does this .qmlls.ini work for providing the correct lsp values?

wallcards/src/LoadingBar.qml
```diff
@@ -0,0 +1,101 @@
...
+        loops: Animation.Infinite
+        property: "value"
+        running: loadingBar.loading
+        target: spinAnimation
+        to: 2 * Math.PI
+      }
+    }
+
+    NText {
+      text: root.pluginApi?.tr("widget.generate-thumbs-message") + String(Math.max(loadingBar.total - loadingBar.pending, 0)).padStart(String(loadingBar.total).length, " ") + " / " + loadingBar.total
```
spiros132: Do prefer to use translation interpolation here, instead of string concatenation.

wallcards/src/ThumbnailService.qml
```diff
@@ -0,0 +1,237 @@
...
+
+  function thumbnailHexValueCmd(thumbnailPath) {
+    return `magick "${thumbnailPath}" \
+      -resize "1x1^" \
+      -gravity center \
+      -extent 1x1 \
+      -depth 8 \
+      -format "%[hex:p{0,0}]" info:- 2>/dev/null \
+      | grep -oE '[0-9A-Fa-f]{6}' \
+      | head -n 1`;
```
spiros132, line 119: I would suggest maybe putting these scripts into their own script files.

wallcards/src/Utils.js
```diff
@@ -10,17 +10,28 @@ function isImage(fileName, filterImages) {
...
+    "--hwdec=auto",
+    "--profile=high-quality",
+    "--video-sync=display-resample",
+    "--interpolation",
+    "--tscale=oversample"
+].join(" ");
+
+function mpvpaperRun(filePath) {
+    return "mpvpaper -o '" + mpvpaperOptions + "' '*' \"" + filePath + "\" >/dev/null 2>&1 & disown";
 }
```
spiros132, line 31: I would suggest instead of implementing your own mpvpaper implementation to maybe use the video-wallpaper plugin in some way. That way both mpvpaper and regular Qt Multimedia video is supported.

wallcards/Wallcards.qml
```diff
@@ -0,0 +1,414 @@
...
+  function applyCurrentCard() {
+    var f = filteredFiles[cardDeck.currentIndex];
+    if (!f)
+      return;
+
+    applicant.command = ["bash", "-c", Utils.wallpaperCommand(f)];
+    applicant.running = true;
+
+    var wallpaperPath = f.isVideo ? f.thumbnail : f.filePath;
+    WallpaperService.changeWallpaper(wallpaperPath);
```
spiros132, line 58: As I wrote before, I would recommend instead of using your own implementation of mpvpaper to use the video-wallpaper plugin.

spiros132 (commented):

wallcards/.qmlls.ini
tonigineer: No, unfortunately not. I just tried some things, but didn't get it to work. Do you know how to?
Will removed it though :)

spiros132: I see, no sadly I don't know either how to fix lsp if noctalia is installed through a package manager, so that's why I was curious if this worked correctly :P

spiros132:
I'll put this into draft for now, do ping me when you're done :D

wallcards/src/Utils.js
tonigineer: Of course, I could do that. Does the plugin dependency in the manifest already work? I remember reading something like "future feature" on the Plugins-Wiki.

I would then use the provided IPC calls or is there an elegant way of importing from another plugin other than using an ugly relative one?

spiros132: Sadly I don't think the dependency field works in the manifest, but I believe that even if it did, you wouldn't want to make it a required dependency, since not all users would use it. 

One suggestion I have is that you put a check at the start to see if the plugin exists and if it doesn't skip all the video related wallcards.

I would suggest the ipc calls, also if any ipc calls that you need are missing, please feel free to add them to that plugin.

tonigineer: Hola, 

I played around with the video-wallpaper plugin and got it to work with IPC calls, but it was not really stable (sometimes freezing, but maybe it was me) and the performance with the Qt Multimedia was poor (browsing cards while having a live paper with Qt). 

Since the currently applied customs calls for `mpvpaper` just work, I would like to keep it that way, since there is not really a benefit without the option to use Qt Multimedia.

Regarding the dependency discussion: I get the opt-out approach, but I wouldn't see any problem with setting another plugin as a dependency :)

So, could we still proceed without implementing the other plugin?

Best regards,
Toni

wallcards/Wallcards.qml
tonigineer: Is already discussed above.

wallcards/src/Utils.js
spiros132: Of course, if that's how you would like your plugin to be I'm completely up for it. As I wrote before, if you feel like the PR is done, put it ready for review and ping me :D

sivansh11:
idk if this is relevant to this current pull request, but I just auto updated now and wallcards started working a bit weirdly.
I have multiple directories of images which I swap using a command 
```
jq --arg dir "$ABS_DIR" '.wallpaper.directory = $dir' "$CONFIG_FILE" > "$CONFIG_FILE.tmp" && mv "$CONFIG_FILE.tmp" "$CONFIG_FILE"
```
on further investigation I noticed that every image in my system is being cached (even those that are explicitly in non of my wallpaper folders).
Note that this doesnt occur with the inbuilt wallpaper selector, it only caches the images accessible in the image directories.

is what I am doing explicitly not supported ? or is this a bug in wallcards ?

tonigineer:
Can definitely be a bug :) I will have a look.

sivansh11:
thanks :D
I am exploring the project trying to pin point the issue, is it okay I tag you here if I find it ? (tho no promises, I dont know qml, so ill be going off intuition)

also sorry for hijacking this pr with my issue, but, are we open to mimicking the image path inside the cache folder ?
so if the user changes the path, wallcards still see the new path specific cached images.
atm its kinda displays everything!

tonigineer:
The problem is, that I changed wallpaper directory to the Noctalia wallpaper dir. The dir in the settings does not work anymore. Don't know about your side effects, but I gonna have a look.

sivansh11:
ummm even before the current update I used to only change noctalia's wallpaper dir, infact prior to this update, there was no dir option in the first place ?

tonigineer:
What I found is, that if you change the directories for the wallpaper, it will show all thumbnails that were ever created. I need to fix this. Is this the problem, that you encounter when changing the directories?

sivansh11:
> What I found is, that if you change the directories for the wallpaper, it will show all thumbnails that were ever created. I need to fix this. Is this the problem, that you encounter when changing the directories? 

Yes I think this is the issue

spiros132:
@tonigineer Are you in the noctalia discord? Somebody is looking for your help there.

PR #765, opened 2026-04-26
add keep-awake-plus plugin

noamsto:
## Summary

Adds **Keep Awake+**, a richer replacement for Noctalia's built-in `KeepAwake` widget.

- Two scopes: `partial` (system stays awake, monitor may sleep) and `full` (everything blocked, uses Noctalia's `IdleInhibitorService`).
- Configurable duration grid (default `30m / 1h / 2h / 4h / 8h`) + optional `∞` unlimited tile.
- Bar pill with scope-tinted icon and optional countdown text.
- One-click `+30m` quick-extend while a session is running, without losing remaining time.
- Persistent last-choice — middle-click the bar widget to re-activate the most recent duration + scope combo.

Backed by a small host-side `system-awake` script (the user's [nix-config](https://github.com/noamsto/nix-config)). The plugin probes for it on PATH at startup; if missing it disables its pollers and shows a single error toast pointing at the source repo, so a misconfigured install fails loudly and gracefully instead of spamming failed processes.

## Plugin folder contents

- `manifest.json` (no `"official"` field — defaults to `false` per registry generator)
- `Main.qml`, `BarWidget.qml`, `Panel.qml`, `Settings.qml`
- `README.md` (features, dependency, settings table, click map)
- `preview.png` (960×540, 16:9)

preview

## Test plan

- [x] Local install via the dev-symlink path (`~/.config/noctalia/plugins/<hash>:keep-awake-plus → repo`)
- [x] Verified: bar pill toggles, scope changes recolor pill, panel duration grid + extend + Turn on/off all work
- [x] Reconfigure (changing scope or duration mid-session) does not flicker the bar widget
- [x] PATH probe path: tested by removing `system-awake` from PATH — single toast fires, pollers stay idle

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Mathew-D:
@noamsto I was looking at your plugin and your comment in the code says to look at https://github.com/noamsto/nix-config for your backend script but I get a 404 error when I went to look. Could the script not be included with plugin in the same folder?

noamsto:
> @noamsto I was looking at your plugin and your comment in the code says to look at [noamsto/nix-config](https://github.com/noamsto/nix-config?rgh-link-date=2026-04-30T00%3A42%3A07Z) for your backend script but I get a 404 error when I went to look. Could the script not be included with plugin in the same folder?

It should absolutely be part of it,  I'm going to fix this

noamsto:
@Mathew-D 
sorry, took me some time.
script is now provided.

Mathew-D:
@noamsto just some friendly input (I have no power):

- you are using console.log in Main.qml where you should be using Logger (Logger.i / Logger.d / Logger.w / Logger.e) 
- can some of you error messages use i18n instead of being hard code
- if the script is going to be in the script folder why does it need to be in the path?  You can just use it from that location
- You might want to read: https://noctalia.dev/blog/announcing-noctalia-v5

noamsto:
Thanks for the friendly input — all three are valid and now addressed in v0.3.0:

- `console.warn` → `Logger.w` in Main.qml (Settings.qml was already on Logger).
- User-facing strings moved to i18n: toasts (enable/disable/extend with scope-based descriptions), the panel header subtitle, the on/off button, and the bar + control-center tooltips.
- The script is now invoked directly from `pluginApi.pluginDir + "/scripts/system-awake"` — no `$PATH` install needed. Removed the PATH probe and the "binary not found" toast, and updated the README accordingly.

Read the v5 announcement too — noted that the whole plugin will need a port when the new plugin API lands. Thanks again!

keep-awake-plus/manifest.json
```diff
@@ -0,0 +1,29 @@
+{
+  "id": "keep-awake-plus",
+  "name": "Keep Awake+",
+  "version": "0.3.0",
+  "minNoctaliaVersion": "3.6.0",
+  "author": "Noam Stolero",
+  "license": "MIT",
+  "repository": "https://github.com/noctalia-dev/noctalia-plugins",
+  "description": "Duration+scope Keep Awake with persistent last-choice and partial/full inhibit modes.",
+  "tags": ["Bar", "ControlCenter", "Power", "KeepAwake"],
```
spiros132: Not all of these tags exist, do check which ones exist at the root README file in this repository.

keep-awake-plus/scripts/system-awake
```diff
@@ -0,0 +1,355 @@
...
+TAG="noctalia-keep-awake"
+SESSION_SUBCMD="__session"
+SCHEMA_VERSION=1
+SELF="$(realpath "$0" 2>/dev/null || echo "$0")"
+
+mkdir -p "$PERSIST_DIR"
+
+notify() {
+  command -v notify-send >/dev/null 2>&1 || return 0
+  notify-send -i "$1" "Keep Awake" "$2" -t 2500 || true
```
spiros132, line 21: These notify-sends don't support multi-language toasts. You don't need to add that if you don't want to, just that it's good to know.

spiros132 (commented):
Some minor feedback, otherwise it's ready to get merged in :D

keep-awake-plus/manifest.json
noamsto: Fixed in `ad0bf916` — switched to `["Bar", "Panel", "System"]` to match the registry tags used by other power/hardware plugins.

keep-awake-plus/scripts/system-awake
noamsto: Plugin always passes `--silent`, so these only fire on direct CLI use. Leaving English for now.

spiros132 (commented):

PR #768, opened 2026-04-26
When you press the alt key, 1 of the axes always remains at 0

adavidys:
to avoid a situation where one extra pixel would shift and change the distances

### DIFF
--- a/screen-toolkit/Measure.qml
+++ b/screen-toolkit/Measure.qml
@@ -252,8 +252,21 @@ Variants {
                 hoverEnabled: true
                 onPositionChanged: (mouse) => {
                     if (overlayWin.measuring) {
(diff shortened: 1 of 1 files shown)


ycf-anon:
@adavidys hey, thanks for the PR 

I’ve made some changes and restructured the plugin (see my PR: https://github.com/noctalia-dev/noctalia-plugins/pull/747). 

It’s just waiting to be reviewed and merged. mesure.qml is now in the overlays folder, so your changes won’t apply as-is.

Could you wait until it’s merged, then open a new PR on top of it? Thanks

@spiros132 does that sound good to you?

adavidys:
If it's just a matter of moving the Measure.qml file to another directory, then I'll submit the PR again in 2 days without any problems.

PR #771, opened 2026-04-26
fix(usb-drive-manager): use standard error hover colors for eject button

Electronic-Mango:
Use standard error colors - `Color.mError` and `Color.mOnError` - for the "eject" button; instead of the "error container" ones. Initially I thought of using the standard "hover" colors, but this should give the "eject" button different color to the rest of them.
This should fix the "eject" button being black when hovering over it.

I'm not sure where the `mErrorContainer` and `mOnErrorContainer` should be from; noctalia-shell doesn't have them defined.


Electronic-Mango:
I updated plugin version to 1.2.1.

PR #776, opened 2026-04-27
feat(usb-drive-manager): add a control center widget

Electronic-Mango:
Add a control center widget to usb-drive-manager plugin, allowing accessing the plugin via a control center shortcut, in addition to regular bar button/widget.

### DIFF
--- a/usb-drive-manager/manifest.json
+++ b/usb-drive-manager/manifest.json
@@ -15,6 +15,7 @@
   "entryPoints": {
     "main": "Main.qml",
     "barWidget": "BarWidget.qml",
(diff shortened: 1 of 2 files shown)


usb-drive-manager/manifest.json
spiros132, line 1: Don't forget to also bump the version number.

spiros132 (commented):
Since this only adds a `ControlCenterWidget` that already uses the functionality of the plugin, I'll wait one more day for @hennifant to review, otherwise I'll merge it in.

Electronic-Mango:
I pushed a version bump to 1.2.0.

PR #778, opened 2026-04-27
Network Indicator - New Features

tonigineer:
Hola, 

added a request for font customization and a horizontal layout for the values. 

Also added a panel showing the network graph from Noctalia's system monitor and a hopefully better preview image :)

Best regards,
Toni


network-indicator/i18n/en.json
spiros132, line 1: There are different lines added and removed from all the translation files, do double check that everything and each translation is included correctly.

spiros132 (commented):
Just double check the translation files, otherwise I'll merge it in :)

network-indicator/i18n/en.json
tonigineer: Thanks for the comment, there was something inconsistent, which should be fixed now.

PR #779, opened 2026-04-27
feat(mirror-mirror) Added center control widget to mirror-mirror

Mathew-D:
As per https://github.com/noctalia-dev/noctalia-plugins/discussions/769
Added a mirror-mirror to allow Control Center.

### DIFF
--- a/mirror-mirror/ControlCenterWidget.qml
+++ b/mirror-mirror/ControlCenterWidget.qml
@@ -1,0 +1,14 @@
+import QtQuick
+import Quickshell
+import qs.Widgets
+
+NIconButtonHot {
+    property ShellScreen screen
+    property var pluginApi: null
+
+    icon: "screen-share"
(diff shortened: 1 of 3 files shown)


xwaleedahmad:
Hi @Mathew-D  — I pulled the latest and I’m definitely running the updated mirror-mirror v0.1.3 (with controlCenterWidget in the manifest), but it still doesn’t show up in Control Center and the plugin throws an error when Control Center tries to load it.

image
image

Mathew-D:
@xwaleedahmad it worked for me but just in case my copy was different I just deleted it, then downloaded and installed the currently version to check.  Try uninstalling it and redownloading maybe something happened to your copy as the ?v=2 isn't anywhere in the code. 
image

xwaleedahmad:
I’d tried everything before this—reinstalling multiple times and even adding the code manually—but nothing helped. Now it’s working perfectly. Really weird, but I’m glad it’s fixed. Thanks!

PR #780, opened 2026-04-27
fix(screen-shot-and-record): Missed a " with the last audio fix

Mathew-D:
I missed remove a " when I removed the audio function.
As per: https://github.com/noctalia-dev/noctalia-plugins/discussions/762


Pulsar33550336:
🤣
I think I should also improve the plugin myself, but I've been busy lately, really no time :D

Mathew-D:
@Ly-sec when you get a sec please.

PR #784, opened 2026-04-28
perf(monique): cache resolved binary path to reduce polling overhead

ToRvaLDz:
## Summary

- **Cache the resolved `monique` binary path at startup** — the `which` lookup runs once during `checkInstalled()`, and the resolved absolute path is reused for all subsequent polling calls (`--current-profile`, `--list-profiles`, `--switch-profile`).
- **Bump default refresh interval from 3s to 5s** — monitor profiles rarely change autonomously, 5s is still responsive enough for hotplug events.

## Problem

On systems using Python version managers like pyenv, the bare `monique` command resolves through a shell shim that spawns ~30 child processes on every invocation (pyenv-exec → pyenv-version-name → pyenv-version-file → pyenv-hooks → pyenv-latest → pyenv-versions → sort → grep → sed → awk → ...).

With the previous 3s polling interval, this resulted in **~600 unnecessary process spawns per minute** just to read a 4-character string (`"Full"`), causing measurable CPU overhead (~8% on the Quickshell process).

## Fix

The `whichProcess` already runs `which monique` at startup. This PR captures its stdout into a new `moniquePath` property and uses it as the command for all subsequent `Process` calls. The shim overhead happens exactly once instead of every 3-5 seconds.

`BarWidget.qml` also uses the cached path from `mainInstance.moniquePath` when opening monique from the context menu.

## Test plan

- [ ] Install monique via system package manager → plugin resolves and caches path correctly
- [ ] Install monique via pip/pipx (user-local) → plugin resolves `~/.local/bin/monique`
- [ ] Uninstall monique → plugin shows "not found" warning, no crashes
- [ ] Profile switching still works via panel
- [ ] Context menu "Open Monique" still launches the GUI
- [ ] Verify reduced CPU usage with `top` or `htop` on systems with pyenv

### DIFF
--- a/monique/BarWidget.qml
+++ b/monique/BarWidget.qml
@@ -85,7 +85,7 @@ Item {
 
   Process {
     id: openMoniqueProcess
-    command: ["monique"]
+    command: [mainInstance?.moniquePath ?? "monique"]
   }
 
   NPopupContextMenu {
--- a/monique/Main.qml
+++ b/monique/Main.qml
@@ -8,9 +8,10 @@ Item {
 
   property var pluginApi: null
 
-  readonly property int refreshInterval: pluginApi?.pluginSettings?.refreshInterval ?? 3000
+  readonly property int refreshInterval: pluginApi?.pluginSettings?.refreshInterval ?? 5000
 
   property bool moniqueInstalled: false
+  property string moniquePath: ""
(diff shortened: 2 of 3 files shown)


spiros132 (commented):
Don't forget to also bump the version number in manifest :)

PR #787, opened 2026-04-28
feat(github-feed): add notification management features and bump the version

pehlicd:
## Changes

Add per-notification and bulk mark as read actions to the notifications tab.

- Hover a notification to reveal a check button that marks it as read and removes it from the list
- "Mark all as read" button in the panel header marks only the currently loaded notifications individually (avoids touching notifications outside the plugin's loaded set)
- Rapid clicks are serialized through a queue

## Motivation

The Notifications tab previously showed unread notifications with no way to act on them without leaving the panel. Users had to open the browser just to dismiss a notification, breaking the workflow.

### DIFF
--- a/github-feed/Main.qml
+++ b/github-feed/Main.qml
@@ -856,6 +856,62 @@ Item {
         }
     }
 
+    function markAllNotificationsAsRead() {
+        if (!root.token || root.notificationsList.length === 0) return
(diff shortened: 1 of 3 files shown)


spiros132 (commented):
Looks good to me, since @linuxmobile hasn't responded in a couple of days and it's a very small change, I'll merge it in :)

PR #788, opened 2026-04-29
feat(usb-drive-manager): add Chinese translations and ptyxis support

PaloMiku:
## Changes

- Add Simplified Chinese (zh-CN) and Traditional Chinese (zh-TW) translations in `i18n/`
- Add `ptyxis` to the terminal emulator dropdown in Settings
- Fix terminal flag detection in `openInFileBrowser()` — ptyxis, gnome-terminal, and wezterm now correctly use `--` instead of `-e`


spiros132:
Would you be able to fix the merge conflicts?

PR #796, opened 2026-04-30
feat(noctalia-calculator): copy result to clipboard from panel

kohanyirobert:
## Summary
- Adds a small copy icon next to the displayed result in the calculator panel; clicking copies the value via `wl-copy` and shows a toast.
- Adds two i18n keys (`panel.copy-result-tooltip`, `toast.copied`) — English only; other locales fall back to English.
- Bumps plugin version to 1.1.0.

## Why
The panel display is plain `Text`, so users cannot select or copy the result with the mouse. This adds the smallest possible discoverable affordance.

## Preview

Calculator panel showing the new copy icon next to the result

## Test plan
- [x] Resting `0` state: icon hidden.
- [x] After `1 + 2 =`: icon visible; click puts `3` in the clipboard (`wl-paste` confirms); toast reads `Copied: 3`.
- [x] Mid-expression (`1 + 2`, no `=`): click copies `2`.
- [x] Error state (`1 / 0 =`): icon hidden.
- [x] Long result (>24 chars): clipboard has full value; toast truncated with `…`.

### DIFF
--- a/noctalia-calculator/Main.qml
+++ b/noctalia-calculator/Main.qml
@@ -2,6 +2,7 @@ import QtQuick
 import Quickshell
 import Quickshell.Io
 import "AdvancedMath.js" as AdvancedMath
+import qs.Services.UI
 
 Item {
     id: root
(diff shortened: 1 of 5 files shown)


spiros132:
@pir0c0pter0

PR #797, opened 2026-04-30
feat(usb-drive-manager): Add "Eject All" to context menu, fix missing icons

Electronic-Mango:
## Add "Eject All" to context menu
Add "Eject All" to context menus, both to bar and control center widgets. Resolves #590
eject

## Fix icon colors when drives are mounted
Use `Color.mPrimary` and `Color.mOnPrimary` for bar widget colors when devices are mounted. Resolves #716

I'm not sure where did the previous `mPrimaryContainer` and `mOnPrimaryContainer` came from, I assume something got updated in noctalia, but not in this plugin. I think the "missing" colors caused the button to be rendered as black.

| Before | After |
|--------|--------|
| before | after |

### DIFF
--- a/usb-drive-manager/BarWidget.qml
+++ b/usb-drive-manager/BarWidget.qml
@@ -51,8 +51,8 @@ NIconButton {
     applyUiScale: false
     customRadius: Style.radiusL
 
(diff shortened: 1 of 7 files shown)


simonjones49:
Thank you

PR #798, opened 2026-04-30
RSS-FEED: Add option to show only when unread

NullSeile:
Adds the option to hide the bar widget when there are no unread posts.


NullSeile:
I was wondering if i should fix these problems, as they are from previous commits.

If you want me to, i can very much do so.

spiros132:
Please do fix the problems if you can :)

NullSeile:
I've fixed the translations!

PR #808, opened 2026-05-02
wallcards - bug fix: creating miniatures

sc0dyx:
Prevent `Cannot create pipe (Too many open files)` error on startup

by replacing the unbounded parallel loop with a job queue limited to 4 simultaneous ffmpeg/magick processes.

Without this, having many wallpaper files (e.g. 40) exhausts system file descriptors and hangs the shell at thumbnail creation.

Details:
- Added maxThumbnailJobs, pendingJobs, activeJobs properties.
- Replaced direct process spawning in createThumbnails() with a queue.
- New functions processNextJob() and createSingleThumbnail() handle recursive dispatch and callback-driven flow.
- Thumbnail generation now runs smoothly even with hundreds of files
2026-05-02-123419_hyprshot

### DIFF
--- a/wallcards/src/ThumbnailService.qml
+++ b/wallcards/src/ThumbnailService.qml
@@ -6,6 +6,9 @@ import Qt.labs.folderlistmodel
 Item {
   id: service
 
+  property int maxThumbnailJobs: 4
+  property var pendingJobs: []
+  property int activeJobs: 0
   required property string cacheDir
   required property var imageFilter
   required property var videoFilter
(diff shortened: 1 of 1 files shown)


tonigineer:
Hola, 

I never ran into the problem on my system (120 files worked fine), but I could replicate the problem with more files. Nevertheless, the initial approach of spawning unlimited processes is bad, so thanks for implementing the "correct" approach. 

Tested your approach on my system and it also worked with multiple hundred files. 

There is still https://github.com/noctalia-dev/noctalia-plugins/pull/807 open from yesterday. I would take care of resolving the conflict, when that one was merged. 

Thanks a lot and best regard,
Toni

tonigineer:
@spiros132 since there are no conflicts, this one can be merged from my side.

PR #809, opened 2026-05-02
feat(network-manager-vpn): Add shortcut to control center

ghuzed:
## Summary
Add a shortcut to Network Manager VPN inside the control center

## Why
Felt like having this personally and decided to share

## What Changed

It's now possible to add Network Manager VPN to the control center's shortcuts. It displays the current connection status and can open the VPN menu on click.

disconnected

connected-hover


spiros132:
Don't forget to bump the version number as well

ghuzed:
Sorry for the delay, found an issue with my previous color logic resulting in no color being used.

PR #814, opened 2026-05-03
Add niri-layout-indicator plugin

alnrog:
Adds Niri Layout Indicator plugin.

A native Noctalia Shell bar widget for niri that displays and switches keyboard layouts.

Features:
- Layout indicator (text / flag)
- Mouse controls (next / previous / menu)
- Context menu with layout selection
- Plugin settings integration


plugins/niri-layout-indicator/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [niri](https://github.com/YaLTeR/niri)
+- Noctalia Shell
+
+---
+## 📦 Installation
+
+```bash
+mkdir -p ~/.config/noctalia/plugins
+cp -r niri-layout-indicator ~/.config/noctalia/plugins/
```
spiros132: This section isn't correct since the user can install the plugin just by pressing install.

plugins/niri-layout-indicator/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+Then enable the plugin in Noctalia settings.
+
+### Quick install
+
+```bash
+./install.sh
```
spiros132: This isn't correct either since there is no `./install.sh` script and there shouldn't be any

plugins/niri-layout-indicator/Settings.qml
```diff
@@ -0,0 +1,567 @@
...
+
+Item {
+  id: rootItem
+  implicitWidth: 600
+  implicitHeight: root.implicitHeight
+  width: Math.max(implicitWidth, parent ? parent.width : 0)
+
+  property var pluginApi: null
+
+  Timer {
```
spiros132: Why is this timer needed in the first place?

plugins/niri-layout-indicator/Settings.qml
```diff
@@ -0,0 +1,567 @@
...
+    implicitWidth: 600
+    width: parent.width
+    spacing: Style.marginM
+
+  // Force minimum width with invisible spacer
+  Item {
+    Layout.preferredWidth: 600
+    Layout.preferredHeight: 0
+    visible: false
+  }
```
spiros132: This `Item` to enforce minimum width shouldn't be needed.

spiros132 (commented):

plugins/niri-layout-indicator/README.md
alnrog: Thanks! Updated the installation instructions to use the Noctalia Plugin Store workflow.

plugins/niri-layout-indicator/README.md
alnrog: Thanks! Removed the Quick install section and switched the installation instructions to the Plugin Store workflow.

plugins/niri-layout-indicator/Settings.qml
alnrog: Good point. The timer was only being used to force the settings popup width. I've removed it since it isn't really related to the plugin functionality itself.

plugins/niri-layout-indicator/Settings.qml
alnrog: Good point. This spacer was only used to enforce a minimum settings width. I've removed it along with the related width workaround.

alnrog:
Thanks again for the review.

I’ve addressed the settings panel feedback:
- removed the resize timer
- removed the invisible spacer used for width forcing
- fixed settings persistence through the Plugin Settings dialog
- added bar widget settings reload so changes apply immediately

The PR is ready for review again.

spiros132 (commented):
Looks good! I'll merge this in :)

PR #821, opened 2026-05-04
Add clamshell plugin

Sovego:
## Summary

  Adds `clamshell`, a Noctalia plugin for automatic laptop clamshell mode on niri.

  When an external display is connected and the plugin is enabled, it starts a single `systemd-inhibit` process to block `handle-lid-switch`. This lets niri
  turn off the internal display on lid close without the system suspending. When no external display is present, or the user disables the plugin, the
  inhibitor is stopped and normal lid behavior returns.

  ## Features

  - Monitors niri output changes through `niri msg --json event-stream`
  - Reactively refreshes `niri msg --json outputs` without polling
  - Detects external displays via configurable internal connector regex
  - Provides Control Center and bar widgets
  - Provides settings for enable state, bar visibility, notifications, internal connector regex, and inhibitor identifier
  - Exposes IPC target `plugin:clamshell`
  - Sends Noctalia toast notifications on inhibitor state changes
  - Does not modify `/etc/systemd/logind.conf`

  ## IPC

  ```sh
  qs -c noctalia-shell ipc call plugin:clamshell status
  qs -c noctalia-shell ipc call plugin:clamshell toggle
  qs -c noctalia-shell ipc call plugin:clamshell enable
  qs -c noctalia-shell ipc call plugin:clamshell disable
  qs -c noctalia-shell ipc call plugin:clamshell refresh
```
  ## Testing

  Tested locally on:

  - niri
  - CachyOS
  - Noctalia Shell / noctalia-qs
  - systemd with systemd-inhibit

  Verified:

  - Plugin loads from local source
  - Settings translations render correctly
  - Control Center and bar widgets load
  - IPC status, toggle, enable, disable, and refresh work
  - External monitor detection works with eDP-1 internal and DP-2 external
  - systemd-inhibit --list shows one noctalia-clamshell inhibitor when active
  - Inhibitor is removed when plugin is disabled
  - Noctalia toast notifications display with icon
  - Static contract test passes:
  -  node tests/static-contract.test.mjs

  ## Notes

  The plugin intentionally uses systemd-inhibit --mode=block and does not change logind configuration. Users should keep normal lid behavior in logind.conf, typically HandleLidSwitch=suspend.

### CODE REVIEW

File: clamshell/manifest.json [hunk 1]
```diff
@@ -0,0 +1,33 @@
+{
+  "id": "clamshell",
```
spiros132, line 2: Since this is specifically for niri, I would suggest renaming the plugin to `niri-clamshell` to make it more clear.

clamshell/manifest.json
```diff
@@ -0,0 +1,33 @@
...
+      "enabled": true,
+      "alwaysShowBarWidget": false,
+      "notify": true,
+      "internalConnectorRegex": "^(eDP|LVDS|DSI)",
+      "inhibitorWho": "noctalia-clamshell"
+    },
+    "ipc": {
+      "target": "plugin:clamshell",
+      "functions": ["enable", "disable", "toggle", "status", "refresh"]
+    }
```
spiros132, line 31: This ipc field in the manifest doesn't do anything.

clamshell/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Installation
+
+Place this directory at:
+
+```sh
+~/.config/noctalia/plugins/clamshell/
+```
+
+Register and enable the plugin in Noctalia, then add the bar widget or control
+center widget from Noctalia settings.
```
spiros132, line 23: The installation part isn't correct since the user can just press install on the plugin.

clamshell/Settings.qml
```diff
@@ -0,0 +1,146 @@
...
+    id: root
+
+    property var pluginApi: null
+    readonly property var main: pluginApi ? pluginApi.mainInstance || ({}) : ({})
+    readonly property var cfg: pluginApi?.pluginSettings || ({})
+    readonly property var defaults: pluginApi?.manifest?.metadata?.defaultSettings || ({})
+
+    property bool editEnabled: cfg.enabled !== undefined
+        ? cfg.enabled
+        : (defaults.enabled !== undefined ? defaults.enabled : true)
```
spiros132, line 16: Prefer to use the following syntax instead of this:
```qml
property bool foo: cfg?.foo ?? defaults?.foo ?? true
```

clamshell/Settings.qml
```diff
@@ -0,0 +1,146 @@
...
+    property bool editAlwaysShowBarWidget: cfg.alwaysShowBarWidget !== undefined
+        ? cfg.alwaysShowBarWidget
+        : (defaults.alwaysShowBarWidget !== undefined ? defaults.alwaysShowBarWidget : false)
+    property bool editNotify: cfg.notify !== undefined
+        ? cfg.notify
+        : (defaults.notify !== undefined ? defaults.notify : true)
```
spiros132, line 22: Prefer to use that syntax on all of these as well.

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+
+    // ───── Plugin API injection ──────────────────────────────────────────
+    // Noctalia's PluginService injects this. Settings are read with the
+    // standard cfg/defaults fallback chain documented in AGENTS.md.
+    property var pluginApi: null
+
+    readonly property var cfg: pluginApi?.pluginSettings || ({})
+    readonly property var defaults: pluginApi?.manifest?.metadata?.defaultSettings
+        || pluginApi?.manifest?.defaultSettings
+        || ({})
```
spiros132, line 28: The defaultSettings will always be at the `pluginApi?.manifest?.metadata?.defaultSettings`, not in the `pluginApi?.manifest?.defaultSettings`

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+
+    readonly property string internalConnectorRegex: cfg.internalConnectorRegex
+        ?? defaults.internalConnectorRegex
+        ?? "^(eDP|LVDS|DSI)"
+    readonly property bool notifyEnabled: cfg.notify !== undefined
+        ? cfg.notify
+        : (defaults.notify !== undefined ? defaults.notify : true)
```
spiros132, line 35: As before, use the other syntax for this property.

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+    readonly property string inhibitorWho: cfg.inhibitorWho
+        ?? defaults.inhibitorWho
+        ?? "noctalia-clamshell"
+
+    // ───── Public state (consumed by widgets via mainInstance) ───────────
+    // `enabled` is the user-facing toggle. Writes are immediately persisted
+    // because widgets and IPC mutate this property directly.
+    property bool enabled: (cfg.enabled !== undefined)
+        ? cfg.enabled
+        : (defaults.enabled !== undefined ? defaults.enabled : true)
```
spiros132, line 45: As before the same thing here. Also I would suggest making this property `readonly` and when you need to change it, change it through the `pluginApi.pluginSettings.enabled` which will update the property automatically.

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+
+    function refresh() {
+        fetchOutputs();
+        return "ok";
+    }
+
+    function applySettings() {
+        root.enabled = (cfg.enabled !== undefined)
+            ? cfg.enabled
+            : (defaults.enabled !== undefined ? defaults.enabled : true);
```
spiros132, line 123: As before this isn't needed if you use the reactive properties of qml instead of manually setting enabled every time.

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+                }
+            }
+        }
+
+        onExited: (exitCode, exitStatus) => {
+            if (exitCode !== 0) {
+                Logger.w(root.logTag, "niri msg outputs exited with code", exitCode);
+            }
+        }
+    }
```
spiros132, line 311: Do you need this `Process` specifically? Quickshell already provides a way to know which monitors exists with the `Quickshell.screens`.

clamshell/Main.qml
```diff
@@ -0,0 +1,513 @@
...
+                return;
+            }
+        }
+
+        // §4.2 fallback: niri-ipc currently has no Output* Event variants.
+        // Re-fetch outputs on any event so real-time hotplug is detected
+        // (niri generates WorkspacesChanged etc. when monitor topology changes).
+        // The debounce timer coalesces rapid bursts into a single fetch.
+        outputsDebounce.restart();
+    }
```
spiros132, line 394: Another thing as well, I believe that noctalia has a built-in way to get the outputs in the `NiriService`. That way you don't need to have more boilerplate code that does the exact same thing as noctalia.

spiros132 (commented):

Sovego:
Thanks for feedback. Its my first PR to such big project) So I fix this issues as soon as possible

PR #825, opened 2026-05-06
fix(video-wallpaper): respect disabled mpvpaper state on startup

theunpleasantowl:
Prior to this commit, Mpvpaper.qml would call activateMpvpaper() unconditionally on startup. This meant that it would activate mpvpaper even if the "Enabled" setting had been toggled off, or even if the user had never selected a wallpaper file.

This commit adds a guard in activateMpvpaper() that returns early when the plugin is disabled or no wallpaper is  selected.


spiros132:
Would you be able to also bump the version number?

PR #833, opened 2026-05-08
Display Settings plugin

fstanis:
# Display Settings

View and modify your display configuration.

## Compatibility

Mainly tested on Niri, but also works on any compositor that supports `wlr-randr` in read-only mode.

On Niri, the plugin allows changing the settings and also parses the persistent KDL configuration (`~/.config/niri/config.kdl`, following any `include` directives) to show the saved output configuration alongside the live state.

## Features

- **Live display info** on every supported compositor via `wlr-randr` (resolution, refresh rate, scale, transform, position, adaptive sync)
- **Editable settings on Niri** - change mode, scale, transform, VRR, and power per output using `niri msg output`
- **Niri saved config** - reads the persistent output configuration from your KDL files, automatically following `include` directives

preview


spiros132 (commented):
Sorry for the late review, it looks great I'll merge this in! :D 

Also in a future update I would suggest also to organize all the components / non-manifest files into their own directory to not have everything in "root".

PR #834, opened 2026-05-09
feat: add battery-monitor-plus plugin

krteke:
## Summary

Add a new `battery-monitor-plus` plugin for Noctalia Shell.

This plugin is based on the official battery widget and panel behavior, with additional battery information and settings.

## Features

- Bar widget using the original battery style
- Optional power draw display in the bar
- Optional remaining time display in the bar
- Battery panel with:
  - status
  - current power draw / charging power
  - remaining time / time until full
  - battery level
  - battery health
  - current power profile
- Settings for:
  - bar power display
  - bar remaining time display
  - refresh interval
  - existing battery widget options
- i18n support:
  - English
  - Simplified Chinese

## Files

- `battery-monitor-plus/manifest.json`
- `battery-monitor-plus/Main.qml`
- `battery-monitor-plus/BarWidget.qml`
- `battery-monitor-plus/Panel.qml`
- `battery-monitor-plus/Settings.qml`
- `battery-monitor-plus/scripts/read_battery_sysfs.sh`
- `battery-monitor-plus/i18n/en.json`
- `battery-monitor-plus/i18n/zh-CN.json`
- `battery-monitor-plus/README.md`
- `battery-monitor-plus/preview.png`

preview


spiros132:
Is there a reason why this is needed compared to the shell's own battery widgets?

NoFilterA1:
I'd personally use this - the built-in battery widget is minimal, while this one exposes health and power draw which is genuinely useful.
One suggestion worth considering: a device switcher in the panel to browse multiple power sources. I use a programmable Corne keyboard with its own battery, and being able to see all devices in one widget (instead of just the main laptop battery) would make this significantly more useful for people with BT/USB peripherals.

Community signal: 👍 from me, hope this gets merged.

krteke:
> Is there a reason why this is needed compared to the shell's own battery widgets?

I believe the built-in battery widget is intentionally designed to remain minimal, and this extension is not meant to replace it.

The purpose of `battery-monitor-plus` is to provide an optional, more information-dense alternative for users who prefer to actively monitor their battery status. It follows the visual style and basic interaction model of the Shell’s built-in battery widget, while adding detailed information that the default widget does not expose.

These additional details may not be useful to every user. However, for users like me who often use a laptop on battery power, battery life and power consumption are important concerns. Having a straightforward way to see power usage and related battery information directly in the status bar is very useful.

krteke:
> I'd personally use this - the built-in battery widget is minimal, while this one exposes health and power draw which is genuinely useful. One suggestion worth considering: a device switcher in the panel to browse multiple power sources. I use a programmable Corne keyboard with its own battery, and being able to see all devices in one widget (instead of just the main laptop battery) would make this significantly more useful for people with BT/USB peripherals.
> 
> Community signal: 👍 from me, hope this gets merged.

Thanks a lot!

That’s a great suggestion. My current use case is mostly around monitoring the laptop’s primary battery, so I may not be able to fully test or cover all external power devices right away, especially BT/USB peripherals with their own batteries.

But I agree that showing multiple power sources in the same widget would be a useful improvement. I’ll keep this in mind and may look into adding a device switcher as a follow-up improvement.

battery-monitor-plus/BarWidget.qml
spiros132, line 1: I'm wondering is there a way to maybe use the already made `BatteryWidget` instead of rewriting the code of it in here? Is there anything that exists in the `BatteryWidget` that there's no way to change here?

spiros132 (commented):
Sorry for the late review, it looks good to me just one tiny question otherwise I'll merge this in :D

battery-monitor-plus/BarWidget.qml
krteke: Good question. I think by BatteryWidget you mean Modules/Bar/Widgets/Battery.qml, right? I don't see a separate BatteryWidget component in the shell tree.

I did reuse the lower-level pieces where possible, mainly NBattery and BatteryService. The reason I didn't import the full built-in Battery.qml directly is that it is a complete bar widget: it owns its settings source, tooltip, context menu, click behavior, and panel/settings routing.

This plugin needs plugin-specific settings, its own panel/settings routing, and the extra power/time text in the bar, so using the built-in widget directly would either keep the original behavior or require shell-side changes.

I agree sharing more code would be nicer long-term. If the shell later exposes a smaller reusable battery content component, I'd be happy to update this plugin to use it.

spiros132: I see, then I understand. Thank you for the PR! I'll merge this in :D

PR #842, opened 2026-05-11
feat(agent-session-status): add Agent session indicator

nzlov:
preview


spiros132:
Do not add the registry file in your PR. That one gets automatically updated when the PR is merged.

nzlov:
> Do not add the registry file in your PR. That one gets automatically updated when the PR is merged.

all ready

agent-session-status/tests/test_server.py
spiros132, line 1: I don't believe this test is needed since the end user won't use it and it will only take away space from their computer.

spiros132 (commented):
Just one thing, otherwise it looks good and I'll merge this in :D

PR #847, opened 2026-05-11
[Arch Updater] Russian translation after changes in version 2.1.6

slserg:


slserg:
change manifest version

PR #849, opened 2026-05-12
[Arch Updater] Russian translation update

slserg:


Astinator:
Thanks for the updated translations!

PR #863, opened 2026-05-16
Add plugin: Password Store (pass)

bojanpy:
Summary

  Adds a launcher provider that integrates with pass (the standard Unix
  password manager). Two search prefixes are supported from the
  launcher:

  - pass <query> — fuzzy search all entries, press Enter to copy the
  password to clipboard
  - po <query> — fuzzy search all entries, press Enter to copy the
  current TOTP code to clipboard

  How it works

  On init the plugin runs find against $PASSWORD_STORE_DIR (falling back
   to ~/.password-store) to build the entry list. Search uses FuzzySort
  from qs.Commons — same engine as the app launcher — so ranking is
  consistent. Activation calls pass -c <entry> or pass otp -c <entry>,
  both of which clear the clipboard after 45 seconds per pass defaults.
  Wayland clipboard (wl-copy) is handled automatically by pass when
  $WAYLAND_DISPLAY is set.

  Dependencies

  - pass — password lookup and clipboard copy
  - pass-otp — TOTP code generation (po prefix)
  - wl-clipboard — Wayland clipboard support (recommended)

  Checklist

  - manifest.json includes id, name, version, minNoctaliaVersion,
  author, license, description, tags
  - minNoctaliaVersion set to 3.6.0 (minimum version with plugin
  launcher provider API)
  - No plugin settings required — works out of the box with a standard
  pass setup
  - README.md with usage, requirements, install instructions, and
  troubleshooting
  - Tested on Hyprland / Wayland


spiros132:
Do not add the registry file, that one gets updated automatically.

bojanpy:
> Do not add the registry file, that one gets updated automatically.

Done :heavy_check_mark:

pass/Main.qml
```diff
@@ -0,0 +1,98 @@
...
+                            provider.passwords = text.trim().split("\n").filter(p => p.length > 0);
+                    }
+                }
+                stderr: StdioCollector {}
+            }
+        }
+    }
+
+    Component.onCompleted: {
+        LauncherProviderRegistry.registerPluginProvider(pluginApi.pluginId, _providerComponent, {});
```
spiros132: Do not add your launcher provider this way, create an object called `LauncherProvider` and provide that in the manifest itself. The plugins aren't the ones responsible with registering and unregistering their launcher providers from / to the shell.

pass/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+  "id": "pass",
```
spiros132: `pass` might be a too small name I would say. Maybe `password-store`, or something similar like that.

pass/README.md
```diff
@@ -0,0 +1,93 @@
...
+# Debian / Ubuntu
+sudo apt install pass pass-extension-otp wl-clipboard
+
+# Fedora
+sudo dnf install pass pass-otp wl-clipboard
+```
+
+## Installation
+
+### Manual Installation
```
spiros132: If this is the only subsection in the `Installation` section, it might confuse users. I would suggest either skip the installation section or add a small part saying that they can just download the plugin from the noctalia settings.

spiros132 (commented):

gs-101:
Can I make feature requests here? I think it would be nice if this also worked with [gopass](https://github.com/gopassw/gopass).

bojanpy:
> Some feedback about the PR! :)

That's a bunch for the feedback! I will incorporate the changes these days and push back :bow:

bojanpy:
> Can I make feature requests here? I think it would be nice if this also worked with [gopass](https://github.com/gopassw/gopass).

@gs-101 For sure, I kinda had this in my mind to be a 0 conf plugin, but if you think we can accomodate both easy then I say go for it!

gs-101:
> > Can I make feature requests here? I think it would be nice if this also worked with [gopass](https://github.com/gopassw/gopass).
> 
> @gs-101 For sure, I kinda had this in my mind to be a 0 conf plugin, but if you think we can accomodate both easy then I say go for it!

Not sure about building the entry list, but I think you can just do it with "(go)pass ls". As for implementing this, only a drop-down menu for "gopass" or "pass" would be necessary, as gopass is compatible with pass commands.

Not gonna lie, I think that even having a "pass" binary on path which is a symlink to "gopass" would be enough. I'll try that.

gs-101:
>Not gonna lie, I think that even having a "pass" binary on path which is a symlink to "gopass" would be enough. I'll try that.

Yep, that was it. And also setting the directory to "/home/YOUR_USER_HERE/.local/share/gopass/stores/root".

The issue is that the plugin doesn't seem to be able to detect age-encrypted secrets. I entered one of my folders and found nothing.

pass/Main.qml
bojanpy: Thank you for the suggestions @spiros132 I tested the implementation and it functions accordingly.

pass/manifest.json
bojanpy: Should be resolved :)

pass/README.md
bojanpy: I did as you said I said install from the plugins page, or clone from the url. I can dumb it down further if you want.

bojanpy:
> > Not gonna lie, I think that even having a "pass" binary on path which is a symlink to "gopass" would be enough. I'll try that.
> 
> Yep, that was it. And also setting the directory to "/home/YOUR_USER_HERE/.local/share/gopass/stores/root".
> 
> The issue is that the plugin doesn't seem to be able to detect age-encrypted secrets. I entered one of my folders and found nothing.

@gs-101 There is a line that says:

        command: [
            "sh", "-c",
            "store=\"${PASSWORD_STORE_DIR:-$HOME/.password-store}\"; " +
            "find \"$store\" -name '*.gpg' | sed \"s|$store/||;s|\\.gpg$||\" | sort"
        ]
        
find *.gpg, the age secrets I think are a different extension, I don't use that encryption but if you want feel free to make a change test if it works on your box, and ill merge it in.

bojanpy:
@gs-101 

Try something like this:

        command: [
            "sh", "-c",
            "store=\"${PASSWORD_STORE_DIR:-$HOME/.password-store}\"; " +
            "find \"$store\" \\( -name '*.gpg' -o -name '*.age' \\) | sed \"s|$store/||;s|\\.gpg$||;s|\\.age$||\" | sort"
        ]

gs-101:
Replacing all the code related to gpg with age did show the entries:

```text
LauncherProvider.qml
170:        "-mindepth", "1", "-type", "f", "-name", "*.age", "-printf", "%P\n"])
174:        "-type", "f", "-name", "*.age", "-printf", "%f\n",
192:      if (!isDir && name.endsWith('.age')) {
```

But pressing Enter on an entry didn't prompt for a password (to unlock the store). Is that intended (sorry if it is, I haven't looked that much into this)?

bojanpy:
> Replacing all the code related to gpg with age did show the entries:
> 
> ```text
> LauncherProvider.qml
> 170:        "-mindepth", "1", "-type", "f", "-name", "*.age", "-printf", "%P\n"])
> 174:        "-type", "f", "-name", "*.age", "-printf", "%f\n",
> 192:      if (!isDir && name.endsWith('.age')) {
> ```
> 
> But pressing Enter on an entry didn't prompt for a password (to unlock the store). Is that intended (sorry if it is, I haven't looked that much into this)?

Gotcha! Ok that will be an issue with gpg too in case pinentry need to be opened to provide a password.

Ive got them unprotected all locally so I never tested. Ill sort that out and you can test again :)

Thanks

PR #725, opened 2026-04-21
feat(keybind-cheatsheet): v3.7.0 — color customization, MangoWC, Niri fixes + Hyprland 0.55 Lua support

blackbartblues:
## Summary

Major release for the **Keybind Cheatsheet** plugin, combining appearance overhaul, MangoWC compositor support, and Niri parser bug fixes.

---

### v3.5.0 — Full per-category color customization + clipboard paste

- **Full per-category color customization:** every key category (Super, Ctrl, Shift, Alt, XF86, Print, Numeric, Mouse, Default, plus description text) now has both a background-color and a text-color picker.
- **Redesigned color picker UI:** replaced the single pencil button with two compact pills per category — left pill shows the background (colored circle + hex), right pill shows text color (circle with letter preview in the selected text color + hex). Pills are sized to the hex-code width; clickability is obvious from hover state so the pencil icon is gone.
- **Clipboard quick-paste:** copy any hex color (`#RRGGBB` or `#RRGGBBAA`) anywhere on your system and a clipboard icon appears in each pill for one-click paste. `wl-paste` is polled every 1.5s while Settings is visible; polling stops on close. A hint line with a clipboard icon above the pickers explains the feature.
- **Theme awareness:** empty bg for modifier keys (Super/Ctrl/Shift) falls back to Noctalia theme accent (`mPrimary`/`mSecondary`/`mTertiary`); empty text override falls back to the global "Key label" color.
- **Panel opacity parity:** panel now follows Noctalia's panel-in-panel opacity rules.
- **Bug fix:** stopped `WARN scene … Unable to assign [undefined] to QColor` caused by a non-existent `Color.mPrimaryContainer` reference.

**Architecture:**
- Extracted reusable `ColorPill.qml` and `ColorPairRow.qml` components.
- Unified all color value properties as `string` so the empty-string fallback sentinel works uniformly.
- Live-preview + revert-on-cancel: edits apply instantly; the snapshot is restored on close if Apply was never pressed.
- 9 new `keyText*` settings + `keyTextDefault` added to `manifest.metadata.defaultSettings`.

---

### v3.5.1 — Niri parser fixes (closes #715)

Addresses all issues reported by @Ito-69 in discussion #715:

- **Fix 1:** Parser no longer follows commented-out includes (`// include "file.kdl"` is skipped).
- **Fix 2:** Auto-discovery of `keybindings.common.kdl` and `keybindings.noctalia.kdl` from the Niri config directory.
[...]

PR #865, opened 2026-05-17
feat(keybind-cheatsheet): Hyprland Lua config support + undescribed-bind editing

blackbartblues:
## Summary

- **Hyprland Lua config (`hyprland.lua`) support.** Hyprland 0.55+ moved from hyprlang `.conf` to a Lua config. Adds a second Hyprland parsing tor that queries `hyprctl binds -j` for the authoritative, already-evaluated bind list — correctly handling `require()`, `for` loops and multi-key chords that static text parsing cannot recover. Categories are recovered by a light scan of `hyprland.lua` + its `require()` modules (`-- N. NAME` headers + `description` literals, with a prefix heuristic for loop-built binds).
- **`.conf` parser kept unchanged as a fallback.** Parser mode `auto`/`lua`/`conf`; `auto` uses the Lua tor when `hyprland.lua` exists, otherwise the existing hyprlang parser. Users on hyprlang are unaffected.
- **Add description / hide for binds without a description.** Undescribed binds are surfaced in a dedicated section instead of being dropped; each can get a custom description inline or be hidden. Overrides are keyed by a stable identity (`submap|modmask|key|flags|dispatcher`, excluding the unstable `__lua` registry ref) so they survive Hyprland restarts. New `showUndescribedBinds` toggle + restore/clear actions in Settings.
- **`refresh` IPC** added so the cheatsheet can be re-parsed from a keybind.
- Fixes a `text` binding loop on the window-height settings field.
- Bumps to **3.6.0**; i18n synced across all 20 languages (English fallback, Polish localized); CHANGELOG updated.

## Test plan

- [x] Lua tor end-to-end verified on Hyprland 0.55.1: categories from `-- N.` headers, loop-built `Workspace 1-5` correctly categorized, multi-key chords resolved
- [x] Undescribed bind (temp `hl.bind` without `description`) → appears in "Without Description"; desc-override moves it out; hidden-override removes it; `showUndescribedBinds=false` hides the section
- [x] `.conf` parser regression: `conf` mode still parses hyprlang configs unchanged
- [x] `qmllint` clean (syntax); widget APIs and icon names verified against noctalia-shell source
- [x] All 21 JSON files valid and structurally identical to `en.json`
- [ ] Reviewer: visual check of inline edit / hide buttons in the panel

### DIFF
--- a/keybind-cheatsheet/CHANGELOG.md
+++ b/keybind-cheatsheet/CHANGELOG.md
@@ -1,5 +1,34 @@
 # Changelog
 
+## [3.6.0] - 2026-05-17
+
+### Features
+
+**Hyprland Lua config support (`hyprland.lua`)**
+- Hyprland 0.55+ replaces hyprlang `.conf` with a Lua config. Added a second Hyprland parsing tor that queries `hyprctl binds -j` for the authoritative, already-evaluated bind list — correctly handling `require()`, `for` loops, and multi-key chords that cannot be recovered by static text parsing
+- Categories are recovered by lightly scanning `hyprland.lua` (and its `require()` modules) for `-- N. NAME` headers and `description = "..."` literals, with a prefix heuristic for loop-generated binds (e.g. `"Workspace " .. i`)
+- Parser mode setting `auto` / `lua` / `conf`: `auto` uses the Lua tor when `hyprland.lua` exists, otherwise the existing `.conf` parser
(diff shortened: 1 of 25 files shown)


blackbartblues:
Consolidated into #725 — the Hyprland 0.55 Lua-config tor + undescribed-bind editing from this PR has been merged on top of the color-customization / MangoWC / Niri work in #725 and shipped together as a single v3.7.0 release. Closing in favor of #725.

PR #873, opened 2026-05-19
Add plugin: Launcher pass

mellotanica:
implemented a launcher provider integrating with gnu pass password manager, it supports both auto typing using wtype and copy to clipboard with wl-copy as well as pass otp

the plugin has been developed with ai assistance but the code has been reviewed as far as I understand quickshell and noctalia plugin API


mellotanica:
ok, I might be happy with its status right now, I won't add any more edits to the first version anymore until I get feedback

launcher-pass/README.md
```diff
@@ -0,0 +1,68 @@
...
+   - **Type <field>**: Type any field using wtype
+
+## Requirements
+
+- [GNU Pass](https://www.passwordstore.org/) password store
+- `wl-copy` for clipboard operations
+- `wtype` for keyboard input (optional)
+- `pass-otp` extension for OTP codes (optional)
+
+## Installation
```
spiros132: This installation section isn't correct since the end user should be able go into settings and install the plugin from there.

launcher-pass/Settings.qml
```diff
@@ -0,0 +1,66 @@
...
+
+  NTextInput {
+    Layout.fillWidth: true
+    label: pluginApi?.tr("settings.storePath.label")
+    description: pluginApi?.tr("settings.storePath.desc")
+    text: root.editStorePath
+    onTextChanged: root.editStorePath = text
+  }
+
+  NTextInput {
```
spiros132: Since this property is a float, it might be wiser to use a slider instead of the user typing the value in.

launcher-pass/Settings.qml
```diff
@@ -0,0 +1,66 @@
...
+    Layout.fillWidth: true
+    label: pluginApi?.tr("settings.typeDelay.label")
+    description: pluginApi?.tr("settings.typeDelay.desc")
+    text: root.editTypeDelay
+    onTextChanged: root.editTypeDelay = text
+  }
+
+  NTextInput {
```
spiros132: Same thing with this one, I would suggest maybe using a slider, or some other type of control to make it easier for the end user.

launcher-pass/LauncherProvider.qml
```diff
@@ -0,0 +1,555 @@
...
+
+  function shellEscape(str) {
+    return str.replace(/'/g, "'\\''")
+  }
+
+  function resolveDir() {
+    return currentPath === "" ? passwordStoreDir : passwordStoreDir + "/" + currentPath
+  }
+
+  function getSetting(key, fallback) {
```
spiros132, line 43: Is there a reason why this is needed? I would suggest instead using qml property bindings. For example.
```qml
readonly property string foo: cfg.foo ?? defaults.foo ?? "Foo"
```

spiros132 (commented):

launcher-pass/Settings.qml
mellotanica: since these two values (this and the previous one) are pretty much fine-tuning stuff and impact on precise timings, I guess a slider would make them harder to set up when actually needed.. how about I hide them inside an "advanced" spoiler or something like that?

mellotanica: well I guess I saw the configuration spoiler widget in a dream, but an "Advanced" tab could work exactly the same..

spiros132 (commented):
Looks good, thank you for the PR! :D

PR #874, opened 2026-05-19
osk-toggle: wvkbd can use flags

Niakr1s:
Enables wvkbd to use flags, like this:
`wvkbd-deskintl -l full,special --landscape-layers full,special`

### DIFF
--- a/osk-toggle/WvkbdBackend.qml
+++ b/osk-toggle/WvkbdBackend.qml
@@ -38,7 +38,7 @@ Item {
     // --- 1. Availability: check binary exists on PATH ---
     Process {
         id: availabilityChecker
-        command: ["sh", "-c", "command -v " + root.wvkbdBinParts[0]]
+        command: ["which", root.wvkbdBinParts[0]]
         onExited: (exitCode, exitStatus) => {
             root.wvkbdOk = exitCode === 0
             if (root.wvkbdOk && !wvkbd.running) preemptChecker.running = true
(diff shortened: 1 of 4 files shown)


spiros132:
Do not add the registry.json file to the PR. That one gets auto-generated

blackteaextract (changes requested):
Thanks! Looks good :)
Could you just remove the registry.json ?

osk-toggle/WvkbdBackend.qml
```diff
@@ -37,7 +38,7 @@ Item {
     // --- 1. Availability: check binary exists on PATH ---
     Process {
         id: availabilityChecker
-        command: ["sh", "-c", "command -v " + root.wvkbdBin]
+        command: ["sh", "-c", "command -v " + root.wvkbdBinParts[0]]
```
blackteaextract: Hey, I also just ran a quick review with Claude and it pointed out:

Now that the `wvkbdBin` setting supports arguments and `wvkbdBinParts[0]` is used to extract just the binary name, the `sh -c` wrapper in `availabilityChecker` has become unnecessary. `command -v` is a shell builtin, which is why `sh` was needed, but `which` is a standalone executable that has identical exit code semantics and can be called directly:

```suggestion
        command: ["which", root.wvkbdBinParts[0]]
```

Niakr1s: Changed that, why not, it works either way.

blackteaextract (approved):
looks good, thanks!
@spiros132

PR #876, opened 2026-05-20
linux-wallpaperengine-controller: v1.3.1

PaloMiku:
## Summary
- **Extract WallpaperColorManager** from Main.qml (~338 lines) — isolates color extraction subsystem with its own Process objects and timers
- **Extract shared WallpaperBadge component** from GridSection.qml and PreviewCard.qml (~300 lines of duplication removed)
- **O(1) wallpaper-by-path lookup** in Panel.qml via Map rebuild on item change
- **Off-screen animation pausing** for AnimatedImage/Video in grid tiles using `isInView` detection
- **Property loading fallback** — restore saved properties from settings.json when `--list-properties` fails or engine is unavailable, instead of losing user config
- **Shell script hardening** — add `set -eu` to all 9 scripts for consistent error handling
- **Minor cleanups** — named timer constants, remove duplicated `formatBytes`, remove unused import

### DIFF
--- a/linux-wallpaperengine-controller/Main.qml
+++ b/linux-wallpaperengine-controller/Main.qml
@@ -2,13 +2,10 @@ import QtQuick
 import Quickshell
 import Quickshell.Io
 
-import "helpers/shared/ColorCacheHelpers.js" as ColorCacheHelpers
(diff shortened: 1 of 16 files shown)


spiros132 (commented):
Looks good to me! I'll merge this in :D

PR #879, opened 2026-05-21
fix(keybind-cheatsheet): repair empty Settings panel + Lua require path resolution (v3.7.2)

blackbartblues:
## Summary

Fixes two regressions reported in #862:

1. **Settings panel rendered empty.** `Settings.qml` and `ColorPairRow.qml` reference sibling components (`ColorPairRow`, `ColorPill`) but Quickshell's `NPluginSettingsPopup` loads them via `Loader.setSource(file://...)` and **does not** add the plugin directory to the QML engine's implicit import list. The engine raised `ColorPairRow is not a type` and aborted the load, leaving the popup empty. Fixed by adding `import "." as Local` to both files and using `Local.ColorPairRow` / `Local.ColorPill`. A plain `import "."` is unsafe here — the plugin's own `Settings.qml` collides with the global `qs.Commons.Settings` singleton (`qmldir defines type as singleton, but no pragma Singleton found in type Settings`). The namespace alias avoids the conflict.

2. **Hyprland Lua: split configs lost their categories.** The Lua parser treated `require("modules.keybinds")` literally as `modules.keybinds.lua` instead of the Lua-standard `modules/keybinds.lua`. For users who split their config into submodules — `hyprland.lua` → `require("modules.keybinds")` → `modules/keybinds.lua` — the file with `-- N. CATEGORY` headers was never read, so `descToCategory` stayed empty and every `hyprctl binds -j` entry fell back to "Other". The `require` handler now converts dots to path separators before resolving (the standard Lua semantic). Existing one-file configs and explicit-path requires (`require("foo/bar")`) are unaffected.

### DIFF
--- a/keybind-cheatsheet/CHANGELOG.md
+++ b/keybind-cheatsheet/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [3.7.2] - 2026-05-21
+
+### Fixes
+
+- **Settings panel was rendering empty**: `Settings.qml` and `ColorPairRow.qml` reference sibling files (`ColorPairRow`, `ColorPill`) but Quickshell's `NPluginSettingsPopup` loads Settings via `Loader.setSource(file://...)` without adding the plugin directory to the implicit import list. The QML engine threw `ColorPairRow is not a type`, aborting the load. Added `import "."` to both files so the sibling components resolve. Reported by @ajgringo619 (#862)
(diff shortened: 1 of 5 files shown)


blackbartblues:
@spiros132 quick ping — this is a regression fix for v3.7.x and affects every user who tries to open the plugin's Settings panel (it currently renders empty for everyone, not just the configs from #862). The second commit in the diff also unblocks @ajgringo619's split Lua config.

All 4 CI checks are green and the diff is small (+29/-15, no behavioral change for working configs). Would appreciate a quick look when you have a moment — happy to address any feedback on the spot.

Tested locally on Hyprland with a fresh noctalia-shell restart after `pkill qs`.

PR #889, opened 2026-05-25
fix(keybind-cheatsheet): theme-aware description color + Mango binds= variant (v3.7.3)

blackbartblues:
## Summary

Two reported regressions/gaps in the keybind-cheatsheet plugin.

### #884 — Niri descriptions invisible on light/transparent backgrounds
- Since v3.5.0 the description text color defaulted to a hardcoded `#E0E0E0` (light gray)
- On light wallpapers shining through a semi-transparent panel, the text is unreadable
- Confirmed against @porcaror's `settings.json` + screenshot: descriptions are present in `cheatsheetData`, just rendered in invisible gray
- **Fix:** manifest default is now empty → `Panel.qml` falls back to `Color.mOnSurface` (theme-aware). `Settings.qml` Reset clears to empty so the theme-aware fallback kicks in
- Existing users with the baked-in `#E0E0E0` should hit Reset on the Description color in Settings → Appearance

### #881 — MangoWC `binds=` (plural, keysym-based) directives not parsed
- MangoWC has two directive families: singular (`bind=`, keycode/QWERTY-mapped) and plural (`binds=`, keysym/layout-aware)
- AZERTY (and other non-QWERTY) users have to use the plural form; the parser only matched the singular one, so the cheatsheet showed no binds
- **Fix:** regex now accepts both forms for `bind`, `axisbind`, and `mousebind`

### DIFF
--- a/keybind-cheatsheet/CHANGELOG.md
+++ b/keybind-cheatsheet/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [3.7.3] - 2026-05-25
+
+### Fixes
+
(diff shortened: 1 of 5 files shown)


blackbartblues:
@spiros132

PR #890, opened 2026-05-27
feat(wallhaven): add Wallhaven wallpaper browser

hakimshifat:
- Add a new Wallhaven plugin for browsing and searching wallpapers from Noctalia Shell.
  - Add a panel with thumbnail grid, refresh, pagination, purity filtering, and wallpaper
  downloads.

  ## Notes

  Requires `curl` or `wget` for downloads. NSFW searches require a Wallhaven API key.


spiros132:
Isn't there a way already to search and set wallpapers from Wallhaven, built in the noctalia shell?

PR #895, opened 2026-05-28
feat(tailscale): add account switching

fstanis:
This PR adds [fast user switching](https://tailscale.com/docs/features/client/fast-user-switching) to the tailscale plugin, specifically a combo box which only appears if you have more than one account (`tailscale switch --list`) and changing its value will switch to that user and refresh the UI.

Useful if you e.g. have separate personal and work networks.

### DIFF
--- a/tailscale/Panel.qml
+++ b/tailscale/Panel.qml
@@ -432,25 +432,19 @@ Item {
             }
           }
 
-          ComboBox {
+          NComboBox {
             Layout.fillWidth: true
             visible: (mainInstance?.accounts?.length ?? 0) >= 2
-            textRole: "name"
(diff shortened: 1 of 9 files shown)


tailscale/Panel.qml
```diff
@@ -432,6 +432,27 @@ Item {
             }
           }
 
+          ComboBox {
```
spiros132: Would you be able to use `NComboBox` instead of the qml provided one?

spiros132 (commented):
Just one tiny thing, otherwise I'll merge this in :D

tailscale/Panel.qml
fstanis: Done

PR #896, opened 2026-05-28
add urgent-on-notification plugin

matt-hb:
Tiny plugin that sets Niri's urgent flag for windows when they receive a notification, so they can use different decorations.


urgent-on-notification/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "urgent-on-notification",
```
spiros132: It looks good just one thing, if this is ONLY for niri with no supported in the future for the other compositors can you rename it with niri in its name? For example: `niri-urgent-on-notification`.

spiros132 (commented):
It looks good, just one tiny thing to check before merging this in :D

urgent-on-notification/manifest.json
matt-hb: Good call, I initially wanted to support hyprland, but since I don't use it and couldn't find a simple way like Niri's IPC, I've renamed the plugin and added a clarification in the description too.

spiros132:
Thank you for the PR! I'll merge it in :D

PR #897, opened 2026-05-28
feat(ntfy-notifications): copy message content on click

zawnk:
This PR adds the small feature that if you click a notification on the notification list, the body of the clicked message will be copied to the clipboard and the panel closes. Sometimes you have errors or other things in there that you want in some other context and before there was no way of copying the contents.


Serph91P (approved):
Looks fine to me

PR #909, opened 2026-06-03
fix(nvibrant): use Quickshell.execDetached() for reliable command execution

NoFilterA1:
## Problem

After the plugin was merged, users reported that clicking the toggle does nothing. The nvibrant command never executes.

## Root cause

`Process.running = true` silently fails to re-execute in Quickshell when the component has already run once. The correct API for fire-and-forget command execution in Quickshell is `Quickshell.execDetached()` — which is what all other plugins in this repository use.

## Fix

- Replace `Process` component + `running = true` with `Quickshell.execDetached()` in both `Main.qml` and `BarWidget.qml`
- Remove now-unused `import Quickshell.Io` imports

### DIFF
--- a/nvibrant/manifest.json
+++ b/nvibrant/manifest.json
@@ -19,7 +19,7 @@
   "metadata": {
     "defaultSettings": {
       "vibranceValue": 512,
-      "displayCount": 1,
+      "displayIndex": 0,
       "enabled": false
     }
   }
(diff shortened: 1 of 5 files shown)


nvibrant/manifest.json
spiros132, line 1: Don't forget to bump the version number

NoFilterA1: yes, thx for remind

PR #937, opened 2026-06-28
fix(keybind-cheatsheet): apply "merge sequential" to hyprctl-sourced binds + feat(keybind-cheatsheet): make Mod2/Mod3/Mod5 first-class customizable keys

drshade:
Two PRs merged:

# Fix: keybind-cheatsheet

## Problem

"Merge sequential keybinds" has no effect on Hyprland setups — consecutive numeric binds (e.g. `Super+1..9`) list individually instead of collapsing into a `Super + 1-9` range, even with the toggle on.

Repro: any Hyprland config where keys aren't statically parseable from the file (a lua / `hl.bind(...)` config, or anything that makes the plugin fall back to `hyprctl binds`). The binds show up correctly grouped and described — they just never merge.

## Fix

In `buildCategoriesFromHyprctl()`:

- attach the already-computed `_mainKey: String(b.key)` and `_mods: mods.join("+")` (plus an empty `_verb`) to each pushed bind, and
- run `mergeSequentialBindsInCategories(categories)` (gated on the `mergeSequentialBinds` setting) before `saveToDb`.

`__lua` binds all share an empty `_verb`, which is fine — `descTemplate` (the description with integers templated) still disambiguates different actions, so unrelated binds won't merge.

## Testing

Hyprland + lua config (`hl.bind` with `description = ...`), workspace binds registered as contiguous `1..10` runs per modifier.

- **Before:** 30 individual `Go to workspace N` / `Move window to workspace N` / `Send window to workspace N` rows.
- **After:** `Super + 1-9`, `Super+Shift + 1-9`, `Super+Alt + 1-9` (workspace 10 lives on key `0`, so it stays a separate row — the merge keys off the integer key, which is correct).

Niri and native-Hyprland-config paths are unchanged.

# Feat: make Mod2/Mod3/Mod5 first-class customizable keys

## What

Adds per-key customization — background colour, text colour, and a new **display-label override** — for the generic X11 modifiers **Mod2 / Mod3 / Mod5**, bringing them in line with every other modifier.

## Why

A key remapped onto Mod3 — e.g. In hyprland `caps:hyper`, which maps Caps Lock to Mod3 (a common "Hyper key" setup) — shows up on the cheat sheet as the literal **`Mod3`**, with no way to colour or rename it. `Mod1`/`Mod4` are already decoded to `Alt`/`Super` in `decodeModmask`, so only Mod2/Mod3/Mod5 were left with unfriendly names.

## Code changes

[...]

PR #938, opened 2026-06-29
feat(keybind-cheatsheet): make Mod2/Mod3/Mod5 first-class customizable keys

drshade:
## What

Adds per-key customization — background colour, text colour, and a new **display-label override** — for the generic X11 modifiers **Mod2 / Mod3 / Mod5**, bringing them in line with every other modifier.

## Why

A key remapped onto Mod3 — e.g. In hyprland `caps:hyper`, which maps Caps Lock to Mod3 (a common "Hyper key" setup) — shows up on the cheat sheet as the literal **`Mod3`**, with no way to colour or rename it. `Mod1`/`Mod4` are already decoded to `Alt`/`Super` in `decodeModmask`, so only Mod2/Mod3/Mod5 were left with unfriendly names.

## Code changes

- New settings `keyColorMod{2,3,5}`, `keyTextMod{2,3,5}`, `keyLabelMod{2,3,5}` (defaults added to `manifest.json`).
- `Panel.qml`: `getKeyColor` / `getKeyTextColor` gain Mod branches; a new `getKeyLabel()` applies the label **at render time**, so it updates live and leaves the cached bind + colour lookup keyed on the original token.
- `ColorPairRow` gains an optional rename field; Mod2/3/5 appear as full rows in **Settings → Appearance**, including the live preview.
- English i18n strings added.

### Naming note

`keyColor*` and `keyText*` are both *colours* in this plugin, so the label override uses a distinct `keyLabel*` namespace rather than overloading `keyText*`.

This works great for me and my cheatsheet looks nice having "Hyper" instead of "Mod3" appear as the modifier through simple customisation of settings.json. Let me know what you think?


drshade:
Combining this into #937 at the maintainer's request — the Mod2/Mod3/Mod5 customization work has been cherry-picked onto that branch, so both changes now ship as a single PR. Closing this one in favour of #937.

PR #943, opened 2026-07-11
ScreenShot & Record: Fixed crashes and added window selection support on Niri

Pulsar33550336:
> ~~I'm back!~~

## Summary

This PR:
1. Fixes a long-standing issue where this plugin could not be used properly on Niri (as it would probabilistically cause noctalia-shell to crash).
The root cause can be traced back to the Qt library, so I cannot fix it. I used grim to capture "bare" screenshots (ppm) and then display them.
This approach even brings faster speed and higher screenshot consistency (you are capturing an image directly, rather than a potentially inconsistent frozen layer), achieving three benefits at once.
2. Added window selection functionality for Niri.
Please note that using this feature has certain prerequisites and limitations, which have been documented in the README and are prompted for reading in the settings interface.
Here is also reiterated:
   1. Window detection requires Niri with [PR #3305](https://github.com/YaLTeR/niri/pull/3305) (not yet merged).
   2. Due to f64 rounding in the PR, window borders may have up to 1px of offset.
   3. Niri builds without `window-geometries` (from #3305) will silently disable window detection — screenshot capture still works normally.
3. Allow silencing most notifications (failure notifications still take effect), and extract the main logic of the screenshot command into a bash script.
4. Allow users to customize the PNG compression level to balance between higher speed and smaller file size.
5. Countless minor tweaks.

## Tests
Already tested under Niri and Hyprland.
Due to device limitations, multi-monitor testing is not possible; testing is welcome.

## Other
Sincere thanks to @Mathew-D for the tremendous contributions during my absence. Without you, my work today would have required much more effort.

### DIFF
--- a/screen-shot-and-record/BarWidget.qml
+++ b/screen-shot-and-record/BarWidget.qml
@@ -23,15 +23,6 @@ NIconButton {
   readonly property string screenName: screen?.name ?? ""
   readonly property real capsuleHeight: Style.getCapsuleHeightForScreen(screenName)
 
-  Component.onCompleted: {
-  }
-
-  Component.onDestruction: {
-  }
-
-  onRecordingChanged: {
-  }
-
   // NIconButton configuration
   baseSize: capsuleHeight
   applyUiScale: false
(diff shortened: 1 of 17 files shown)


Pulsar33550336:
~~A bug in Hyprland. Blocking.~~

Fixed in 803d25f9

Pulsar33550336:
I noticed some checks didn't run for some reason, so I re-triggered them by reopening.

PR #956, opened 2026-08-13
cloudflare-warp: fix ToS flag, fix mode display, add mode switcher

LycorisND:
## Problem

`warp-cli` refuses every command with *"Please accept the WARP Terms of Service by running this command in a TTY or by passing the --accept-tos flag"* when it isn't run in a TTY. QuickShell's `Process` always runs non-interactively, so `status`, `connect`, and `disconnect` were silently failing on every single invocation — the widget could never show a correct connected state, and clicking connect/disconnect did nothing but log a warning.

The mode display was also dead code: it tried to regex-match a `Daemon Mode:` line out of `warp-cli status` output, but that line doesn't exist in the actual output, so `warpMode` was always an empty string and the mode row never rendered.

## Changes

- Add `--accept-tos` to all `warp-cli` invocations (`status`, `connect`, `disconnect`).
- Fetch the current mode from `warp-cli --json settings` (`operation_mode` field) instead of parsing nonexistent text output.
- Add a mode switcher: a dropdown in the panel covering all `warp-cli mode` values (`warp`, `doh`, `warp+doh`, `dot`, `warp+dot`, `proxy`, `tunnel_only`), calling `warp-cli mode <mode>` on selection.
- Widen the panel (300 → 420) and the mode combo box (min-width 340) so the longer mode names aren't clipped.
- Add `en`/`ru` translations for the new mode names and a "mode changed" toast.

## Testing

Manually exercised the full cycle (`status` → `disconnect` → `status` → `connect` → `status`, plus `mode doh`/`mode warp` switches) via the same `warp-cli` invocations the plugin now uses, confirmed against `qs -c noctalia-shell log` that the plugin loads and refreshes with no warnings/errors.

### DIFF
--- a/cloudflare-warp/Main.qml
+++ b/cloudflare-warp/Main.qml
@@ -16,6 +16,7 @@ Item {
   property string warpMode: ""
   property bool isRefreshing: false
   property bool isSwitchingMode: false
+  property bool modeSwitchLocked: false
   property string lastToggleAction: ""
 
   readonly property var availableModes: [
@@ -136,6 +137,7 @@ Item {
         try {
           var data = JSON.parse(output)
           root.warpMode = data?.settings?.operation_mode ?? ""
+          root.modeSwitchLocked = data?.settings?.switch_locked ?? false
         } catch (e) {
           Logger.w("CloudflareWarp", "Failed to parse warp-cli settings JSON: " + e)
           root.warpMode = ""
@@ -143,6 +145,7 @@ Item {
       } else {
         Logger.w("CloudflareWarp", "warp-cli settings failed: " + String(modeProcess.stderr.text || "").trim())
       }
+      root.isRefreshing = false
     }
   }
 
--- a/cloudflare-warp/Main.qml
+++ b/cloudflare-warp/Main.qml
@@ -15,7 +15,12 @@ Item {
   property bool warpConnected: false
   property string warpMode: ""
   property bool isRefreshing: false
+  property bool isSwitchingMode: false
   property string lastToggleAction: ""
+
+  readonly property var availableModes: [
+    "warp", "doh", "warp+doh", "dot", "warp+dot", "proxy", "tunnel_only"
+  ]
 
   Timer {
     id: updateTimer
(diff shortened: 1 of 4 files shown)


ToRvaLDz (approved):
Thanks for this, really solid catch on the ToS flag, that bug made the whole widget non-functional in non-TTY contexts. I pulled the branch and tested it locally against `warp-cli 2026.6.880.0` in a running noctalia-shell instance. Summary: **works cleanly, approving.** 👍

**Verified locally ✅**
- `--accept-tos` on all invocations, correct and necessary; `Process` runs non-interactively so this was failing on every call.
- The mode-key set (`warp`, `doh`, `warp+doh`, `dot`, `warp+dot`, `proxy`, `tunnel_only`) matches `warp-cli mode <MODE>` exactly.
- `warp-cli --json settings` → `data.settings.operation_mode` is the right path, and the value serializes as the CLI-style string (e.g. `"warp+doh"`), so it maps correctly onto `currentKey`. Confirmed against a live daemon.
- `Panel.qml` loads with no QML errors, the `NComboBox` API usage is accepted by the component.

**One suggestion 🔧, guard against locked mode switching**
On Teams/managed deployments the daemon can forbid mode changes (`warp-cli settings mode-switch-allowed` → `false`, also exposed as `switch_locked` in the JSON you already parse). Right now the combobox stays enabled there, so every selection fails with an error toast. Since you're already reading the settings JSON, you can wire this up with no extra command:

```qml
// Main.qml, alongside warpMode
root.modeSwitchLocked = data?.settings?.switch_locked ?? false
```
```qml
// Panel.qml, on the NComboBox
enabled: !(mainInstance?.isSwitchingMode ?? false) && !(mainInstance?.modeSwitchLocked ?? false)
```

**Optional nits (non-blocking)**
- `refresh()` now starts both `statusProcess` and `modeProcess`, but `isRefreshing` is only cleared in `statusProcess.onExited`, introduced here since `modeProcess` is new. Harmless in practice (setting `running = true` on an already-running `Process` is a no-op), just slightly asymmetric; clearing it in `modeProcess.onExited` too would tidy it up.
- `operation_mode` can also report `posture_only`, which has no combobox entry, it would fall through to the placeholder. Fine to leave as-is, worth a comment.
- 300 → 420 is a fairly large jump right after some panel-width tuning; shortening the longer mode labels would be an alternative to widening. Your call.

Great contribution, thanks again. 🙏

